Issue #840, opened 2021-10-26
Moving to async infra: Tried to access clusterControl on ActorSystem and it was nil


ktoso:
Since moving to the more async infra we sometimes hit this:

```

21:03:52 DistributedActors/ActorSystem.swift:439: Fatal error: BUG! Tried to access clusterControl on ActorSystem(second, sact://second@127.0.0.1:9011) and it was nil! Please report this on the issue tracker.
21:03:55 Current stack trace:
21:03:55 0    libswiftCore.so                    0x00007fa1b96bb060 swift_reportError + 50
21:03:55 1    libswiftCore.so                    0x00007fa1b97335e0 _swift_stdlib_reportFatalErrorInFile + 112
21:03:55 2    libswiftCore.so                    0x00007fa1b94338cc <unavailable> + 1419468
21:03:55 3    libswiftCore.so                    0x00007fa1b94336f3 <unavailable> + 1418995
21:03:55 4    libswiftCore.so                    0x00007fa1b9432200 _assertionFailure(_:_:file:line:flags:) + 411
21:03:55 5    swift-distributed-actorsPackageTests.xctest 0x00005560ddf9bc70 <unavailable> + 5426288
21:03:55 6    swift-distributed-actorsPackageTests.xctest 0x00005560ddf9e619 <unavailable> + 5436953
21:03:55 7    swift-distributed-actorsPackageTests.xctest 0x00005560ddf9e7e0 <unavailable> + 5437408
21:03:55 8    libswiftCore.so                    0x00007fa1b944b660 DefaultStringInterpolation.appendInterpolation<A>(_:) + 23
21:03:55 9    swift-distributed-actorsPackageTests.xctest 0x00005560ddf9bb99 <unavailable> + 5426073
21:03:55 10   swift-distributed-actorsPackageTests.xctest 0x00005560ddf9e619 <unavailable> + 5436953
21:03:55 11   swift-distributed-actorsPackageTests.xctest 0x00005560ddf9e7e0 <unavailable> + 5437408
21:03:55 12   libswiftCore.so                    0x00007fa1b944b660 DefaultStringInterpolation.appendInterpolation<A>(_:) + 23
21:03:55 13   swift-distributed-actorsPackageTests.xctest 0x00005560ddf9bb99 <unavailable> + 5426073
21:03:55 14   swift-distributed-actorsPackageTests.xctest 0x00005560ddf9e619 <unavailable> + 5436953
[...]

PR #852, opened 2021-11-02
=build update swift syntax dependency a little bit

ktoso:
Update swift syntax a bit.


ktoso:
Failure in 5.6 was https://github.com/apple/swift-distributed-actors/issues/840, otherwise ok.

PR #853, opened 2021-11-02
Update readme with more info on build setup

ktoso:
Follow up based on https://forums.swift.org/t/are-there-notes-or-docs-on-how-to-use-a-nightly-development-snapshot-with-projects-swift-distributed-actors/53170/10

Thanks @slashmo for the info on CLion, I can confirm it works great!

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,9 +115,6 @@ export TOOLCHAIN=/Library/Developer/Toolchains/swift-DEVELOPMENT-SNAPSHOT-2021-1
 
 # Just build the project
 $TOOLCHAIN/usr/bin/swift build --build-tests
-
-# Build and run all tests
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -120,16 +120,18 @@ $TOOLCHAIN/usr/bin/swift build --build-tests
 $TOOLCHAIN/usr/bin/swift test
```
heckj, line 120: I recommend clipping this out as well - if you're reading straight down, it looks like the stock `swift test` is expected to work, even with detailing more the the DYLD details just below.

heckj (commented):
Lookin' great, thank you!!

README.md
ktoso: Great catch, thank you!

Issue #855, opened 2021-11-04
Use ManagedAtomicLazyReference in ActorSystem


ktoso:
We have a few properties that are difficult and async to initialize and we use a complex dance to do so.

Today they're either lock protected, or not protected at all. Search for `    // TODO: Use ManagedAtomicLazyReference to store this`

All those should become wrapped like:

```
    private let _receptionistStore: ManagedAtomicLazyReference<OpLogDistributedReceptionist>
    public var receptionist: DistributedReceptionist {
        guard let value = _receptionistStore.load() else {
            fatalError("Somehow attempted to load system.receptionist before it was initialized!")
        }

        return value
    }
```

This way we can never have memory unsafe behavior there, but it'll at least be well defined.

The initialization dance must be taken care of still

PopFlamingo:
@ktoso The issue wasn't automatically closed weirdly

ktoso:
Thanks, closed now!

PR #857, opened 2021-11-09
WIP: Use ManagedAtomicLazyReference in ActorSystem

PopFlamingo:
Use ManagedAtomicLazyReference in ActorSystem for better race condition handling, see #855

### DIFF
--- a/Sources/DistributedActors/ActorSystem.swift
+++ b/Sources/DistributedActors/ActorSystem.swift
@@ -100,20 +100,37 @@ public final class ActorSystem: _Distributed.ActorTransport, @unchecked Sendable
     // ==== ----------------------------------------------------------------------------------------------------------------
     // MARK: Metrics
 
-    // TODO: Use ManagedAtomicLazyReference to store this
-    private lazy var _metrics: ActorSystemMetrics = ActorSystemMetrics(self.settings.metrics)
-    internal var metrics: ActorSystemMetrics {
-        self._metrics
-    }
+    internal let metrics: ActorSystemMetrics
 
     // ==== ----------------------------------------------------------------------------------------------------------------
     // MARK: Cluster
+    internal class Box<T> {
+        var value: T
+        init(_ value: T) {
+            self.value = value
+        }
+    }
 
     // initialized during startup
     // TODO: Use ManagedAtomicLazyReference to store this
     internal var _cluster: ClusterShell?
-    internal var _clusterControl: ClusterControl?
-    internal var _nodeDeathWatcher: NodeDeathWatcherShell.Ref?
+
+    private let _clusterControlStore: ManagedAtomicLazyReference<Box<ClusterControl>>
+    public var cluster: ClusterControl {
+        guard let box = _clusterControlStore.load() else {
+            fatalError("Somehow attempted to load system.cluster before it was initialized!")
+        }
+        return box.value
+    }
+
+    internal let _nodeDeathWatcherStore: ManagedAtomicLazyReference<Box<NodeDeathWatcherShell.Ref?>>
+    internal var _nodeDeathWatcher: NodeDeathWatcherShell.Ref? {
+        guard let box = _nodeDeathWatcherStore.load() else {
+            fatalError("Somehow attempted to load system.nodeDeathWatcher before it was initialized!")
+        }
+        return box.value
+    }
+    
 
     // ==== ----------------------------------------------------------------------------------------------------------------
     // MARK: Logging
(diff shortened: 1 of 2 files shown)


Sources/DistributedActors/ActorSystem.swift
```diff
@@ -100,20 +100,37 @@ public final class ActorSystem: _Distributed.ActorTransport, @unchecked Sendable
...
+
+    private let _clusterControlStore: ManagedAtomicLazyReference<Box<ClusterControl>>
+    public var cluster: ClusterControl {
+        guard let box = _clusterControlStore.load() else {
+            fatalError("Somehow attempted to load system.cluster before it was initialized!")
+        }
+        return box.value
+    }
+
+    internal let _nodeDeathWatcherStore: ManagedAtomicLazyReference<Box<NodeDeathWatcherShell.Ref?>>
```
PopFlamingo, line 126: I would have preferred to keep this private but I think it is needed for `_ActorShell.swift` since it needs to determine wether or not the property has been set (which it used to do using the `nil` value of the optional.

Sources/DistributedActors/ActorSystem.swift
```diff
@@ -100,20 +100,37 @@ public final class ActorSystem: _Distributed.ActorTransport, @unchecked Sendable
...
+        var value: T
+        init(_ value: T) {
+            self.value = value
+        }
+    }
 
     // initialized during startup
     // TODO: Use ManagedAtomicLazyReference to store this
     internal var _cluster: ClusterShell?
-    internal var _clusterControl: ClusterControl?
```
PopFlamingo, line 115: AFAICT this was only needed by the `cluster` property so I removed it

Sources/DistributedActors/ActorSystem.swift
```diff
@@ -100,20 +100,37 @@ public final class ActorSystem: _Distributed.ActorTransport, @unchecked Sendable
...
+    internal let metrics: ActorSystemMetrics
 
     // ==== ----------------------------------------------------------------------------------------------------------------
     // MARK: Cluster
+    internal class Box<T> {
+        var value: T
+        init(_ value: T) {
+            self.value = value
+        }
+    }
```
PopFlamingo, line 112: Please tell me if this would be better at global scope on some utils.swift file for instance

Sources/DistributedActors/ActorSystem.swift
```diff
@@ -250,7 +270,7 @@ public final class ActorSystem: _Distributed.ActorTransport, @unchecked Sendable
 
             initializationLock.withWriterLockVoid {
```
PopFlamingo, line 271: Am I right that this is not needed anymore?

Sources/DistributedActors/_ActorShell.swift
```diff
@@ -159,7 +159,7 @@ public final class _ActorShell<Message: ActorMessage>: _ActorContext<Message>, A
         self._log = Logger.make(system.log, path: address.path)
 
         self.supervisor = _Supervision.supervisorFor(system, initialBehavior: behavior, props: props.supervision)
-        self._deathWatch = DeathWatchImpl(nodeDeathWatcher: system._nodeDeathWatcher ?? system.deadLetters.adapted())
+        self._deathWatch = DeathWatchImpl(nodeDeathWatcher: system._nodeDeathWatcherStore.load()?.value ?? system.deadLetters.adapted())
```
PopFlamingo, line 162: As far as I can tell this is actually acceptable for the value not to be set yet here because the type where this is indirectly used (`LocalActorRefProvider`) is invoked before the set in the init dance.

Sources/DistributedActors/ActorSystem.swift
PopFlamingo: Thinking about it I think it is actually still needed for `_cluster`

ktoso:
Awesome, thanks! I'll give it a deeper look shortly :)

ktoso:
(nightly CI will fail now, we'll update CI to a known toolchain and then run)

Sources/DistributedActors/ActorSystem.swift
ktoso: It can check this based on if cluster is enabled or not I believe :) I'll give it a poke

Sources/DistributedActors/_ActorShell.swift
ktoso: Hmmm... the shell always needs the deathwatch, I believe... I'll double check this; perhaps the only actor we start so early is one which does never deathwatch so it's fine... 🤔 

I'll double check but this PR is fine!

ktoso (approved):

ktoso:
Checked locally and this is good, thank you!

CI is on nightly which moved too far and fails our build now -- we'll get back a stable CI shortly, sorry for "living on the edge" issues 🙏

Sources/DistributedActors/ActorSystem.swift
```diff
@@ -100,20 +100,37 @@ public final class ActorSystem: _Distributed.ActorTransport, @unchecked Sendable
...
+    internal final class Box<T> {
+        var value: T
+        init(_ value: T) {
+            self.value = value
+        }
+    }
 
     // initialized during startup
     // TODO: Use ManagedAtomicLazyReference to store this
     internal var _cluster: ClusterShell?
```
ktoso, line 116: That's another one we'd love to be not the variable anymore, do you want to take it up?

PR #860, opened 2021-11-27
Store ClusterShell using a ManagedAtomicLazyReference

PopFlamingo:
This is the continuation of what was done in #857 (see also #855) : `ManagedAtomicLazyReference` is used to protect the `_cluster: ClusterShell` property as well against memory level race conditions.

### DIFF
--- a/Sources/DistributedActors/ActorSystem.swift
+++ b/Sources/DistributedActors/ActorSystem.swift
@@ -112,8 +112,13 @@ public final class ActorSystem: _Distributed.ActorTransport, @unchecked Sendable
     }
 
     // initialized during startup
-    // TODO: Use ManagedAtomicLazyReference to store this
-    internal var _cluster: ClusterShell?
+    private let _clusterStore: ManagedAtomicLazyReference<Box<ClusterShell?>>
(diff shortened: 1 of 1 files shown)


Sources/DistributedActors/ActorSystem.swift
```diff
@@ -447,10 +449,12 @@ public final class ActorSystem: _Distributed.ActorTransport, @unchecked Sendable
 
             /// Only once we've shutdown all dispatchers and loops, we clear cycles between the serialization and system,
             /// as they should never be invoked anymore.
+            /*
             self.lazyInitializationLock.withWriterLockVoid {
                 // self._serialization = nil // FIXME: need to release serialization
-                self._cluster = nil
             }
```
PopFlamingo, line 455: Note tant this whole section is now commented out

ktoso:
Excellent, thank you :-)

ktoso:
I think we're done with https://github.com/apple/swift-distributed-actors/issues/855 for now even 🤔 

resolves https://github.com/apple/swift-distributed-actors/issues/855

PR #863, opened 2021-12-01
invoking docc from toolchains

### LINKED ISSUE #856: Since DocC now ships in toolchains, adjust scripts/docs/generate... to not have to download/build docc
We used to have to pull, build and use the in-place build docc since it did not ship with the toolchain.

Now it does, so we can adjust the scripts -- they can just invoke `docc` available in the toolchains 👍 

More info: https://forums.swift.org/t/swift-docc-is-now-included-in-swift-toolchains-for-both-macos-and-linux/53284/2

shilpeegupta14:
Adjusted the generate_docc.sh file to invoke docc from toolchains now.

### Motivation:

Now that the DocC ships in toolchains, I adjusted the generate_docc.sh file to invoke docC directly from toolchains. I am making this change in order to learn contributing effectively. I am trying to update the DocC.

### Modifications:
1) changed the prepare and build toolchain(earlier line no 50 to 73) code to directly access docc from toolchains.
2) modified and wrote
     export TOOLCHAIN=/Library/Developer/Toolchains/swift-DEVELOPMENT-SNAPSHOT-2021-11-20-a.xctoolchain.  
     cd $TOOLCHAIN/usr/bin/docc

### Result:

- Resolves #856 
Screenshot 2021-12-01 at 6 54 03 PM

### DIFF
--- a/scripts/docs/generate_docc.sh
+++ b/scripts/docs/generate_docc.sh
@@ -47,30 +47,13 @@ declare -r build_path_linux="$build_path/"
 declare -r docc_source_path="$root_path/.build/swift-docc"
 declare -r docc_render_source_path="$root_path/.build/swift-docc-render"
 
-# Prepare and build docc
-if [[ ! -d "$docc_source_path" ]]; then
-  git clone https://github.com/apple/swift-docc.git "$docc_source_path"
-  cd $docc_source_path
-
-  if [[ ! -d "$docc_source_path/$build_path" ]]; then
-    swift build -c release
-  fi
-else
-  echo "Assuming docc is built..."
+# invoking docc from toolchain
+if [-z ${TOOLCHAIN+x} ]; then   
+  echo "TOOLCHAIN is unset"; exit(1);
 fi
+export DOCC_HTML_DIR=${TOOLCHAIN}/usr/bin/docc
 
-
-if [[ ! -d "$docc_render_source_path" ]]; then
-  git clone https://github.com/apple/swift-docc-render.git "$docc_render_source_path"
-  cd $docc_render_source_path
-
-  npm install
-  npm run build
-else
-  echo "Assuming docc-render is built..."
-fi
-
-export DOCC_HTML_DIR=$docc_render_source_path/dist
+$TOOLCHAIN/usr/bin/docc
 
 # Build documentation
 


scripts/docs/generate_docc.sh
```diff
@@ -47,30 +47,9 @@ declare -r build_path_linux="$build_path/"
...
-
-  npm install
-  npm run build
-else
-  echo "Assuming docc-render is built..."
-fi
-
-export DOCC_HTML_DIR=$docc_render_source_path/dist
+# Accessing docc from toolchain
+export TOOLCHAIN=/Library/Developer/Toolchains/swift-DEVELOPMENT-SNAPSHOT-2021-11-20-a.xctoolchain
```
ktoso, line 51: I mean, no, we cannot hardcode the script to a specific path -- it should ask for the toolchain to be set instead for example -- require the TOOLCHAIN var exists.

scripts/docs/generate_docc.sh
```diff
@@ -47,30 +47,9 @@ declare -r build_path_linux="$build_path/"
...
+cd $TOOLCHAIN/usr/bin/docc
```
ktoso, line 52: this is a binary, we don't "cd" it.

scripts/docs/generate_docc.sh
shilpeegupta14: Something like this....

if  [[ ! -d "$docc_source_path" ]]; then 
  cd /Library/Developer/Toolchains/swift-DEVELOPMENT-SNAPSHOT-2021-11-20-a.xctoolchain
else
  echo "TOOLCHAIN var exists."

scripts/docs/generate_docc.sh
shilpeegupta14: okay so only this much will help
$TOOLCHAIN/usr/bin/docc

scripts/docs/generate_docc.sh
heckj: @shilpeegupta14 - I suspect what you want to do is check to see if the environment variable TOOLCHAIN is already set, and if so - use it. For a linux platform with a nightly swift pre-installed and already added to the relevant paths (such as a docker build from a nightly swift toolchain release), then you won't need to prefix the commands with TOOLCHAIN, as they *should* be already on the path (I haven't double-checked that hypothesis though).

In either case, you'll want to likely set the `DOCC_HTML_DIR` environment variable so that `docc`, however you invoke it, knows where to find its HTML goodies it wants to combine/link with the output.

[StackOverflow has a nice reference](https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash) to this otherwise awkward problem of determining if an environment variable is already set, which amounts to doing something like:
`if [ -z ${var+x} ]; then echo "var is unset"; else echo "var is set to '$var'"; fi`

Translating that to "require an environment TOOLCHAIN to be set in order to proceed" might be:

`if [ -z ${TOOLCHAIN+x} ]; then echo "TOOLCHAIN is unset"; exit(1); fi`

Then from there, build up the `DOCC_HTML_DIR`:

`export DOCC_HTML_DIR=${TOOLCHAIN}/usr/share/docc/render/`

scripts/docs/generate_docc.sh
heckj: yep, invoke with `$TOOLCHAIN/usr/bin/docc` once you've verified that TOOLCHAIN is defined.

shilpeegupta14:
Adjusted the generate_docc.sh file to invoke docc from toolchains now.

**Motivation:**

Now that the DocC ships in toolchains, I adjusted the generate_docc.sh file to invoke docC directly from toolchains. I am making this change in order to learn contributing effectively. I am trying to update the DocC.

**Modifications:**

1) changed the prepare and build toolchain(earlier line no 50 to 73) code to directly access docc from toolchains.
2) modified and wrote
```
 if [ -z ${TOOLCHAIN+x} ]; then 
    echo "TOOLCHAIN is unset"; exit(1); 
 fi
 export DOCC_HTML_DIR = ${TOOLCHAIN}/usr/bin/docc

```
3) `$TOOLCHAIN/usr/bin/docc`
Screenshot 2021-12-04 at 9 51 16 AM

shilpeegupta14:
@swift-server-bot please test

scripts/docs/generate_docc.sh
```diff
@@ -47,30 +47,13 @@ declare -r build_path_linux="$build_path/"
...
-  git clone https://github.com/apple/swift-docc-render.git "$docc_render_source_path"
-  cd $docc_render_source_path
-
-  npm install
-  npm run build
-else
-  echo "Assuming docc-render is built..."
+# invoking docc from toolchain
+if [-z ${TOOLCHAIN+x} ]; then   
+  echo "TOOLCHAIN is unset"; exit(1);
```
Kyle-Ye, line 52: This gives an error on my machine with zsh.

line 52: syntax error near unexpected token `1'

scripts/docs/generate_docc.sh
```diff
@@ -47,30 +47,13 @@ declare -r build_path_linux="$build_path/"
...
 fi
+export DOCC_HTML_DIR=${TOOLCHAIN}/usr/bin/docc
 
-export DOCC_HTML_DIR=$docc_render_source_path/dist
+$TOOLCHAIN/usr/bin/docc
```
Kyle-Ye, line 56: Calling docc seems useless here for me.

What we want is to export the env DOCC_HTML_DIR so that we can use it later when call docc in preview_docc.sh

scripts/docs/generate_docc.sh
```diff
@@ -47,30 +47,13 @@ declare -r build_path_linux="$build_path/"
...
-
-  npm install
-  npm run build
-else
-  echo "Assuming docc-render is built..."
+# invoking docc from toolchain
+if [-z ${TOOLCHAIN+x} ]; then   
+  echo "TOOLCHAIN is unset"; exit(1);
 fi
+export DOCC_HTML_DIR=${TOOLCHAIN}/usr/bin/docc
```
Kyle-Ye, line 54: DOCC_HTML_DIR should point to the docc-render dist dir or in the toolchain it is located at /usr/share/docc/render

ktoso:
Replaced by https://github.com/apple/swift-distributed-actors/pull/865

Thanks for the effort @shilpeegupta14, maybe you'll find something else you can contribute; 
Thanks @Kyle-Ye for the PR!

PR #867, opened 2022-02-02
Down replaced cluster members

### LINKED ISSUE #866: Gossip re-adds nodes replaced after ungraceful shutdown
Commit hash: `dcd618bcf93b60bdf19ca1b29c24c4a08d30615a`

(HEAD of `release/0.x`, behavior should be same on HEAD of `main` branch)

Context:

If a node in an actors cluster is killed ungracefully (say, via `SIGKILL`) and its replacement is immediately respawned then the previous node is still considered to be `up` since SWIM has not yet reached consensus on the previous node being `down` yet. In this scenario, a call to `Cluster.Membership.removeCompletely()` will be made [when processing the `.joining` MembershipChange](https://github.com/apple/swift-distributed-actors/blob/662e08172ca54e2a43da84381b5b07fe3fe6a070/Sources/DistributedActors/Cluster/Cluster%2BMembership.swift#L304) for the replacement node which will immediately remove it as a cluster member. However, if the cluster subsequently receives a gossip update that still includes the previous node as `.up` then the `Cluster.Membership.mergeFrom()` function [will create a MembershipChange directive](https://github.com/apple/swift-distributed-actors/blob/662e08172ca54e2a43da84381b5b07fe3fe6a070/Sources/DistributedActors/Cluster/Cluster%2BMembership.swift#L527) to add the node back.

Subsequently, the replacement node gets marked `.down` by the downing strategy (unclear exactly how this happens) while the previous node stays in the cluster membership marked as `.up`. This persists after additional restarts.

Steps to reproduce:

1. Run any actors cluster with 3 nodes using a static IP + port combination for each node
2. `kill -9` the process for one node
3. Immediately restart that node using the same IP + port

budde:
Make sure a previous cluster member is marked as `.down` before replacing it

### Motivation:

* Resolve #866 

### Modifications:

* When processing a `MemberChange` that replaces an existing node, mark the node as `.down` if it hasn't been already
* Add an integration test for #866 (confirmed failure before this change.

### Result:

- Resolves #866

### DIFF
--- a/IntegrationTests/tests_04_cluster/it_Clustered_swim_ungraceful_shutdown/main.swift
+++ b/IntegrationTests/tests_04_cluster/it_Clustered_swim_ungraceful_shutdown/main.swift
@@ -1,0 +1,54 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Distributed Actors open source project
+//
+// Copyright (c) 2022 Apple Inc. and the Swift Distributed Actors project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.md for the list of Swift Distributed Actors project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import DistributedActors
+
+var args = CommandLine.arguments
+args.removeFirst()
+
+guard args.count >= 1, let bindPort = Int(args[0]) else {
+    fatalError("Bind port must be provided")
+}
+
+print("Binding to port \(bindPort)")
+
+let system = ActorSystem("System") { settings in
+    settings.logging.logLevel = .info
+
+    settings.cluster.enabled = true
+    settings.cluster.bindPort = bindPort
+
+    settings.cluster.swim.probeInterval = .milliseconds(300)
+    settings.cluster.swim.pingTimeout = .milliseconds(100)
+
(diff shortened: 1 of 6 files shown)


budde:
Also have a `down-replaced-cluster-members-0.x` in my fork against the `release/0.x` branch

ktoso:
Whoa that's a nasty compiler crash there... Thank you for looking into this I will have a look!
CI is weird since it is 5.6 maybe we should actually use 5.5 on this branch... Will think about it.

ktoso:
Change looks good to me! I'll reproduce and play around a bit more tomorrow -- can you submit the PR to 0.x and let's leave main be for the time being? I'll forward port the patch later then 👍

ktoso:
Minor typo in the integration test :)

```
16:29:43 /code/IntegrationTests/tests_04_cluster/test_02_ungraceful_shutdown_recovery.sh: line 31: thrid_logs: unbound variable

```

ktoso:
Hmmm the integration test is not aggressive enough perhaps since it does not reproduce the issue but I think the fix looks 
right...

budde:
Good catch, was making a few silly assumptions in the integ test. Think the typo you caught might've been contributing to why I was getting it to fail initially ;) 

Updated the test. Should fail deterministically without the fix.

ktoso:
Great, reproduced the failure and confirmed the fix -- looking good, thank you a lot @budde !

I'll cut a release for this as well :)

PR #871, opened 2022-03-16
0.x CI setup

tomerd:
* add docker setup for 5.6
* update 5.4 to use release version


ktoso (approved):
LGTM, thanks -- thinking if we should just merge as we know the 5.6 will continue to be borked for a while

PR #876, opened 2022-04-03
Adapt codebase for Swift 5.6 release version

segabor:
Adapt codebase for Swift 5.6 release version

### Motivation:

I wanted to try out Distributed Actors library on my Mac having Swift 5.6 release installed but it could not even compile. So I decided to fix the code base. 

### Modifications:

I applied the following fixes to the codebase,
- Fixed and updated SwiftSyntax dependency in Package.swift
- Fixed SwiftSyntax import in GenActions Swift source
- Made a test class final to meet a requirement imposed by Actorable protocol.

### Result:

The whole codebase now compiles with Swift 5.6. Also the full test suite is passed. All the changes made on a MacBook Pro mid 2015 running the latest macOS (Monterey 12.3.1) and Xcode 13.3.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -68,7 +68,7 @@ var targets: [PackageDescription.Target] = [
         name: "GenActorsLib",
         dependencies: [
             "DistributedActors",
-            .product(name: "SwiftSyntax", package: "SwiftSyntax"),
+            .product(name: "SwiftSyntaxParser", package: "SwiftSyntax"),
             .product(name: "Logging", package: "swift-log"),
             .product(name: "ArgumentParser", package: "swift-argument-parser"),
 
@@ -307,7 +307,7 @@ dependencies.append(
 )
 #elseif swift(>=5.6)
 dependencies.append(
-    .package(name: "SwiftSyntax", url: "https://github.com/apple/swift-syntax.git", .exact("0.50600.0"))
+    .package(name: "SwiftSyntax", url: "https://github.com/apple/swift-syntax.git", .exact("0.50600.1"))
 )
 #elseif swift(>=5.5)
 dependencies.append(
--- a/Package.swift
+++ b/Package.swift
@@ -18,6 +18,12 @@ var globalConcurrencyFlags: [String] = []
 // ])
 // #endif
 
+#if swift(>=5.6)
+var swiftSyntaxDep="SwiftSyntaxParser"
+#else
+var swiftSyntaxDep="SwiftSyntax"
+#endif
+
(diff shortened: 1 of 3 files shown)


tomerd:

segabor:
Not sure if this is something that can be easily fixed but my knowledge is quite limited. So, `SwiftSyntaxParser` type has been extracted to a separate product but the binary release for Swift 5.4.x does not contain this change.

There are multiple ways it can be fixed,

1. Ship new `SwiftSyntax` release for older Swift versions and update `Package.swift` to include them
2. Hack around `Package.swift` and make sure to include the right dependency (`SwiftSyntax` vs `SwiftSyntaxParser`) based on what's in the binary dep.
3. Drop support of older Swift versions

segabor:
@tomerd can I get an automated test please?

ktoso:
The release/0.x branch can require Swift 5.6, that's okey -- no need for backwards compatibility.
All our adopters are able to move on to it quickly.

ktoso:

Package.swift
```diff
@@ -18,6 +18,12 @@ var globalConcurrencyFlags: [String] = []
 // ])
 // #endif
 
+#if swift(>=5.6)
+var swiftSyntaxDep="SwiftSyntaxParser"
+#else
+var swiftSyntaxDep="SwiftSyntax"
+#endif
```
ktoso, line 25: - formatting is off needs space between the name and `=`
- please declare those as `let`, we never mutate them

ktoso:
Yeah it is known that this didn't right now build on 5.6 because issues with SwiftSyntax.
There was also a blocker issue in the syntax release itself, but that got solved.

If you're interested I'd really welcome the help on getting us green here on this branch and 5.4 and 5.6.
If we can't make it work on 5.4 anymore we could consider dropping it tbh as this isn't a stable release branch.

ktoso:
Oh actually, seems everything is fixed with Syntax now, thank you :-) I'll fix the formatting issue and we can merge 🥳

ktoso:

Package.swift
ktoso: I went ahead and fixed this since otherwise the patch is LGTM  -- thank you very much!

segabor:
@ktoso happy to help!

ktoso:
And we're green :-)

Thanks for the poke here, I was recently head down on the main branch and updating us to Swift 5.7 and some final compiler work to make it possible, please look forward to the main branch becoming stable very soon! 🥳

Issue #879, opened 2022-05-17
Rename ActorSystem since it gets confusing with the ActorSystem typealias in Swift DA


ktoso:
* ActorSystem → ClusterSystem
* Might need to rename or get rid of <system.cluster> since maybe folks will now store it as let cluster = ClusterSystem()

This is started on prepare-5.7 just need to go all the way

yim-lee:
Discussed with @ktoso, for this issue, we will:

- [ ] Rename `ActorSystem` -> `ClusterSystem` (rename file also)
- [ ] Merge `ClusterSettings` into `ClusterSystemSettings` (rename file also)

We will not rename `system.cluster` but will revisit later in https://github.com/apple/swift-distributed-actors/issues/892.

The `enabled` flag in `ClusterSettings` should be removed. That work is covered by https://github.com/apple/swift-distributed-actors/issues/880.

ktoso:
That all sounds great, thank you for looking into these!

ktoso:
Awesome, thank you!

PR #893, opened 2022-05-18
Rename ActorSystem to ClusterSystem

yim-lee:
Part 1 of https://github.com/apple/swift-distributed-actors/issues/879

### DIFF
--- a/Sources/DistributedActors/LifecycleMonitoring/LifecycleWatch.swift
+++ b/Sources/DistributedActors/LifecycleMonitoring/LifecycleWatch.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift Distributed Actors open source project
 //
-// Copyright (c) 2018-2021 Apple Inc. and the Swift Distributed Actors project authors
+// Copyright (c) 2018-2022 Apple Inc. and the Swift Distributed Actors project authors
 // Licensed under Apache License v2.0
 //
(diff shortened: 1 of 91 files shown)


yim-lee:
Rename of `ActorSystemSettings` will be done in a separate PR.

ktoso (approved):
LGTM, skimmed through it all and looks good!

There was quite a few small related types with old names still... Thank you a lot for spotting them all :)

Sources/DistributedActors/LifecycleMonitoring/LifecycleWatch.swift
```diff
@@ -26,7 +26,7 @@ public extension LifecycleWatch {
     @discardableResult
     func watchTermination<Watchee>(
         of watchee: Watchee,
-        @_inheritActorContext @_implicitSelfCapture whenTerminated: @escaping @Sendable(ActorSystem.ActorID) -> Void,
+        @_inheritActorContext @_implicitSelfCapture whenTerminated: @escaping @Sendable(ID) -> Void,
```
ktoso, line 29: oh that looks silly... formatters don't understand that all so well I guess;

Technically `@Sendable (ID) -> Void` and not `@Sendable(ID) -> Void` but I don't mind, let's get the bigger PR in and we can always reformat later :)

ktoso:
Awesome, everything LGTM!

PR #899, opened 2022-05-24
[test] remove dead letter test which is hard to reproduce in Swift DA

ktoso:
This test is a bit weird in the new world -- it does not make sense in the local only world, because if we have a reference to an actor it IS alive, so we'll never deadletter here.

The same can be reproduced cross node, but we test that in other ways I believe.


ktoso:
Just integration tests need a revamp now still

Issue #880, opened 2022-05-17
Always enable clustering


ktoso:
* remove settings.cluster.enabled
* no-one will be using this library for “just concurrency” anymore
* if any tests were doing “just concurrency stuff” we can probably get rid of those

PR #902, opened 2022-05-25
Always enable clustering

yim-lee:
Resolves https://github.com/apple/swift-distributed-actors/issues/880


yim-lee:
Seeing `Fatal error: Somehow attempted to load system.cluster before it was initialized!` in different tests and not sure how to track down the cause.

ktoso:
> Seeing Fatal error: Somehow attempted to load system.cluster before it was initialized! in different tests and not sure how to track down the cause.

I've seen that a few times and track it as https://github.com/apple/swift-distributed-actors/issues/840

It is about the actor starting "early" before the cluster var is initialized; this should not ever happen, so I'll look into it. If you see that failure ignore it for now

Tests/DistributedActorsTests/Cluster/MembershipGossipLogicSimulationTests.swift
```diff
@@ -19,10 +19,6 @@ import NIO
 import XCTest
 
 final class MembershipGossipLogicSimulationTests: ClusteredActorSystemsXCTestCase {
-    override func configureActorSystem(settings: inout ClusterSystemSettings) {
-        settings.enabled = false // not actually clustering, just need a few nodes
-    }
```
ktoso, line 24: Eh Oh Interesting... I guess maybe we do want to allow for this just of our internal tests then... :-\ 

Will think some more.

ktoso:
Actually let's merge -- and I'll fix one test after this

Tests/DistributedActorsTests/TraversalTests.swift
```diff
@@ -82,7 +82,9 @@ final class TraversalTests: ActorSystemXCTestCase {
                 "system",
                 "receptionist",
                 "receptionist-ref",
-                // "gossip", // TODO: should it also be here?
+                // cluster actors ---
+                "downingStrategy", "swim", "nodeDeathWatcher", "gossip", "cluster", "leadership",
+                // end of cluster actors ---
```
ktoso, line 87: this test needs to account for the cluster actors now 👍 

Seems all others do pass, at least locally for me

ktoso (approved):
This is LGTM, I'll merge this in a moment, just need to phase merges

Tests/DistributedActorsTests/TraversalTests.swift
yim-lee: 🙏

Issue #881, opened 2022-05-17
Configurable ask timeout for all remoteCalls


ktoso:
When making a `remoteCall` (and `remoteCallVoid`) (in `ActorSystem`) we set a timeout for the `Ask`.

- This must be checking a task-local for the timeout for this specific call
- The default value must be a setting (`self.settings.remoteCallTimeout`)

PR #903, opened 2022-05-25
Make remote call timeout configurable

yim-lee:
Resolves https://github.com/apple/swift-distributed-actors/issues/881

### DIFF
--- a/Sources/DistributedActors/ClusterSystem.swift
+++ b/Sources/DistributedActors/ClusterSystem.swift
@@ -889,7 +889,7 @@ public extension ClusterSystem {
         let recipient = _ActorRef<InvocationMessage>(.remote(.init(shell: clusterShell, address: actor.id._asRemote, system: self)))
 
         let arguments = invocation.arguments
-        let ask: AskResponse<Res> = recipient.ask(timeout: RemoteCall.timeout ?? self.settings.callTimeout) { replyTo in
+        let ask: AskResponse<Res> = recipient.ask(timeout: RemoteCall.timeout ?? self.settings.defaultRemoteCallTimeout) { replyTo in
             let invocation = InvocationMessage(
                 targetIdentifier: target.identifier,
                 arguments: arguments,
@@ -918,7 +918,7 @@ public extension ClusterSystem {
         let recipient = _ActorRef<InvocationMessage>(.remote(.init(shell: shell, address: actor.id._asRemote, system: self)))
 
         let arguments = invocation.arguments
-        let ask: AskResponse<_Done> = recipient.ask(timeout: RemoteCall.timeout ?? self.settings.callTimeout) { replyTo in
+        let ask: AskResponse<_Done> = recipient.ask(timeout: RemoteCall.timeout ?? self.settings.defaultRemoteCallTimeout) { replyTo in
             let invocation = InvocationMessage(
                 targetIdentifier: target.identifier,
                 arguments: arguments,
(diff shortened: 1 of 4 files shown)


Sources/DistributedActors/ClusterSystem.swift
```diff
@@ -889,7 +889,7 @@ public extension ClusterSystem {
         let recipient = _ActorRef<InvocationMessage>(.remote(.init(shell: clusterShell, address: actor.id._asRemote, system: self)))
 
         let arguments = invocation.arguments
-        let ask: AskResponse<Res> = recipient.ask(timeout: .seconds(5)) { replyTo in
+        let ask: AskResponse<Res> = recipient.ask(timeout: RemoteCall.timeout ?? self.settings.callTimeout) { replyTo in
```
yim-lee, line 892: There is already `self.settings.callTimeout`, which based on the description is for this purpose. I can rename this to `remoteCallTimeout` if we prefer that.

yim-lee:

Sources/DistributedActors/ClusterSystem.swift
ktoso: The rename sounds good -- `remoteCallTimeout` or maybe `defaultRemoteCallTimeout` even.

Sounds good 👍 

Is the value there also pretty large, like 5 sec?

ktoso:
Can we add a unit test that sets the value e.g. to 1 second and we get an `AskError` out of the call then?

Something like:

```swift
    func test_remoteCallTimeout_shouldConfigureTimeout() async throws {
        let worker = Worker()

        let error = try shouldThrow { // TODO: this will need to gain an async version in ShouldMatchers.swift
            RemoteCall.withTimeout(.seconds(1)) {
                _ = try await worker.hello()
            }
        }

        guard case AskError.timedOut(let timeoutError) = error else {
            throw testKit.fail("Expected AskError.timedOut, got \(error)")
        } 

        // TODO: assert the timeoutError.timeout == 1.second
    }
```

ktoso:
Looks good! would like to add the test :)

Sources/DistributedActors/ClusterSystem.swift
```diff
@@ -930,6 +930,11 @@ public extension ClusterSystem {
 
         _ = try await ask.value // discard the _Done
     }
+
+    enum RemoteCall {
```
ktoso, line 934: (move to top-level please), and maybe some minor docs like:

```suggestion
/// Allows for configuring of remote calls by setting task-local values around a remote call being made.
/// 
/// ### Example: Override remote call timeouts
/// ```
/// try await RemoteCall.with(timeout: .seconds(1)) { 
///     try await greeter.greet("Caroline")
/// }
/// ```
public enum RemoteCall { // does swiftformat remove that public? grrr
```

I guess we need that "nicer API" as well rather than just exposing the task local hmmm...

ktoso (approved):

Sources/DistributedActors/ClusterSystem.swift
yim-lee: `defaultRemoteCallTimeout` it is

Tests/DistributedActorsTests/ClusterSystemTests.swift
```diff
@@ -149,4 +150,40 @@ final class ClusterSystemTests: ActorSystemXCTestCase {
...
+        let local = await setUpNode("local") { settings in
+            settings.enabled = true
+        }
+        let remote = await setUpNode("remote") { settings in
+            settings.enabled = true
+        }
+        local.cluster.join(node: remote.cluster.uniqueNode)
+
+        let greeter = DelayedGreeter(actorSystem: local)
+        let remoteGreeter = try DelayedGreeter.resolve(id: greeter.id, using: remote)
```
yim-lee, line 164: Can the test setup be simplified?

ktoso (approved):

PR #911, opened 2022-05-27
[beta-2] remove children concept

ktoso:
WIP


ktoso:
This was largely replaced with the actor metadata work, replacing children will wait

Issue #891, opened 2022-05-17
Remove or adapt WorkerPool to distributed actors


ktoso:
It is not really used by us so we can drop it.

We can bring it back when its reimplemented as distributed actor working with distributed actors.

Maybe it is simple enough to reimpl; but we can just remove as well for now

PR #912, opened 2022-05-27
[WIP] Convert WorkerPool to distributed actor

yim-lee:
Resolves https://github.com/apple/swift-distributed-actors/issues/891


yim-lee:
⚠️ Still need to update tests. Want to get some feedback first.

Sources/DistributedActors/Pattern/WorkerPool.swift
```diff
@@ -12,18 +12,28 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Distributed
 import Logging
 
-/// A `WorkerPool` represents a pool of actors that are all equally qualified to handle incoming messages.
+public protocol DistributedWorker: DistributedActor {
```
ktoso, line 18: I'm on the fence if we should lock it down to the actor system:

```suggestion
public protocol DistributedWorker: DistributedActor where ActorSystem == ClusterSystem {
```

My gut feeling is that yes, that might be better for now.

ktoso: (There's an entire world of pain related to the "we need a concrete ActorSystem at all times" because of some language limitations -- by locking it down right away we'll avoid bumping into weird issues)

Sources/DistributedActors/Pattern/WorkerPool.swift
```diff
@@ -12,18 +12,28 @@
...
 ///
 /// Pool members may be local or remote, // TODO: and there should be ways to say "prefer local or something".
 ///
 /// A pool can populate its member list using the `Receptionist` mechanism, and thus allows members to join and leave
 /// dynamically, e.g. if a node joins or removes itself from the cluster.
 ///
 // TODO: A pool can be configured to terminate itself when any of its workers terminate or attempt to spawn replacements.
-public struct WorkerPool<Message: ActorMessage> {
-    typealias Ref = WorkerPoolRef<Message>
+public distributed actor WorkerPool<Worker: DistributedWorker>: DistributedWorker, LifecycleWatch, CustomStringConvertible where Worker.ActorSystem == ClusterSystem {
```
ktoso, line 33: If we'd add the where clause to the DistributedWorker protocol it could likely go away from here 🤔

Sources/DistributedActors/Pattern/WorkerPool.swift
```diff
@@ -44,19 +54,79 @@ public struct WorkerPool<Message: ActorMessage> {
         /// The worker pool will terminate itself if all of its static workers have terminated.
         /// You may death-watch the worker pool in order to react to this situation, e.g. by spawning a replacement pool,
         /// or gracefully shutting down your application.
-        case `static`([_ActorRef<Message>])
+        case `static`([Worker])
```
ktoso, line 57: Hmm... we might need to be careful here and _not_ store the `[Worker]` as they'd never be able to be removed by terminating then 🤔

Sources/DistributedActors/Pattern/WorkerPool.swift
ktoso: Maybe you're right here though... keeping the protocol "wide open" but constraining when used with the pool... 🤔 

(Sorry, just thinking out loud -- just an option to consider if you'd hit some issues with this protocol and calling stuff on it)

Sources/DistributedActors/Pattern/WorkerPool.swift
```diff
@@ -44,19 +54,79 @@ public struct WorkerPool<Message: ActorMessage> {
...
     }
 
     // ==== ------------------------------------------------------------------------------------------------------------
     // MARK: State shared across all states of the WorkerPool
 
-    var settings: WorkerPoolSettings<Message>
+    var settings: WorkerPoolSettings<Worker>
     var selector: Selector {
         self.settings.selector
     }
```
ktoso, line 66: So since we store the selector, we'd keep the `.static[Worker]` as well, making them not able to terminate...

Maybe we should not store `Selector` but some `_Selector` that is same as Selector but for `static` is just `case static` rather than keeping the actors retained 🤔

Sources/DistributedActors/Pattern/WorkerPool.swift
```diff
@@ -44,19 +54,79 @@ public struct WorkerPool<Message: ActorMessage> {
...
-
-    init(settings: WorkerPoolSettings<Message>) {
+    
+    /// The worker pool. Worker will be selected round-robin.
+    private var workers: [Weak<Worker>] = []
```
ktoso, line 69: 👍

Sources/DistributedActors/Pattern/WorkerPool.swift
```diff
@@ -44,19 +54,79 @@ public struct WorkerPool<Message: ActorMessage> {
...
+
+        switch self.selector {
+        case .dynamic(let key):
+            Task {
+                for await worker in await self.actorSystem.receptionist.subscribe(to: key) {
+                    system.log.log(level: self.settings.logLevel, "Got listing member for \(self.selector): \(worker)")
+                    self.workers.append(Weak(worker))
+                    watchTermination(of: worker) { self.onWorkerTerminated(id: $0) }
+                }
+            }
```
ktoso, line 87: need to store the task in the actor, and as we `deinit` cancel it (`task.cancel()`)

Sources/DistributedActors/Pattern/WorkerPool.swift
```diff
@@ -44,19 +54,79 @@ public struct WorkerPool<Message: ActorMessage> {
...
+            workers.forEach { worker in
+                watchTermination(of: worker) { self.onWorkerTerminated(id: $0) }
+            }
+        }
+    }
+    
+    public distributed func submit(work: WorkItem) async throws -> WorkResult {
+        let worker = try await self.selectWorker()
+        return try await worker.submit(work: work)
+    }
```
ktoso, line 99: nice 👍

Sources/DistributedActors/Pattern/WorkerPool.swift
```diff
@@ -44,19 +54,79 @@ public struct WorkerPool<Message: ActorMessage> {
...
+    
+    private func selectWorker() async throws -> Worker {
+        // Wait if we haven't received the initial workers listing yet.
+        // Otherwise, the pool has become empty because all workers have been terminated,
+        // in which case we either wait for new workers or throw error.
+        while self.workers.isEmpty {
+            switch (self.hasTerminatedWorkers, self.settings.whenAllWorkersTerminated) {
+            case (false, _), (true, .awaitNewWorkers):
+                try await Task.sleep(nanoseconds: 100)
```
ktoso, line 108: yeah... So I'd hope we can avoid the sleep loop here with a continuation to wake up when we get a new actor discovered and we'd await on that continuation here

Sources/DistributedActors/Pattern/WorkerPool.swift
```diff
@@ -44,19 +54,79 @@ public struct WorkerPool<Message: ActorMessage> {
...
+            case (true, .crash(let error)):
+                throw error
+            }
+        }
```
ktoso, line 112: Probably also some max timeout for making the call here, so we should stop after callTimeout perhaps and throw from here?

Sources/DistributedActors/Pattern/WorkerPool.swift
```diff
@@ -44,19 +54,79 @@ public struct WorkerPool<Message: ActorMessage> {
...
+
+        if let worker = self.workers[self.roundRobinPos].actor {
+            self.roundRobinPos = (self.roundRobinPos + 1) % self.workers.count
+            return worker
+        } else {
+            // Clean up and try again
+            self.hasTerminatedWorkers = true
+            self.workers.removeAll { $0.actor == nil }
```
ktoso, line 120: okey 👍 Probably possible to do somehow more efficiently in future but this is okey for now -- perhaps we'd store them in a dictionary by their `ID` and then know which one to remove?

Sources/DistributedActors/Pattern/WorkerPool.swift
```diff
@@ -44,19 +54,79 @@ public struct WorkerPool<Message: ActorMessage> {
...
+            self.roundRobinPos = 0
+            return try await self.selectWorker()
+        }
+    }
+    
+    private func onWorkerTerminated(id: ID)  {
+        for worker in self.workers where worker.actor?.id == id {
+            worker.actor = nil
```
ktoso, line 128: this should not be necessary since the terminated can only ever happen if the actort was released.
If it was released -> our Weak<> no longer holds it as well; We should though remove the specific actor from our `workers` then though.

I'd suggest moving workers to `[Worker.ID: Weak<Worker>]` this way we can then immediately `workers.remove(forKey: id)` 👍

ktoso (commented):
This is shaping up good I think, thank you for getting through it so quickly :)

It is a bit different than the previous one since our capabilities changed a little bit with the reliance on ARC now, but nice to see we can still do it 👍

Sources/DistributedActors/Pattern/WorkerPool.swift
yim-lee: Amended https://github.com/apple/swift-distributed-actors/pull/912/commits/28f02afd8ffa218d856c26e69fb7f6b7cd71aa14

Sources/DistributedActors/Pattern/WorkerPool.swift
yim-lee: The pool relies on `ClusterSystem` features such as receptionist, logger, etc. so I suppose we do need the constraint?

Sources/DistributedActors/Pattern/WorkerPool.swift
yim-lee: Yeah that's what I had to begin with but the compiler wasn't happy.

Tried this just now:

```
/code/Sources/DistributedActors/Pattern/WorkerPool.swift:83:60: error: generic parameter 'Guest' could not be inferred
                for await worker in await self.actorSystem.receptionist.subscribe(to: key) {
                                                           ^
/code/Sources/DistributedActors/Cluster/Reception/OperationLogDistributedReceptionist.swift:346:29: note: in call to function 'subscribe(to:)'
    public nonisolated func subscribe<Guest>(
                            ^
/code/Sources/DistributedActors/Pattern/WorkerPool.swift:97:17: error: instance method 'watchTermination(of:whenTerminated:file:line:)' requires the types 'Worker.ActorSystem' and 'ClusterSystem' be equivalent
                watchTermination(of: worker) { self.onWorkerTerminated(id: $0) }
```

Sources/DistributedActors/Pattern/WorkerPool.swift
yim-lee: Is it redundant to watchTermination?

Sources/DistributedActors/Pattern/WorkerPool.swift
yim-lee: Not storing the `[Worker]` or `Selector`.

Sources/DistributedActors/Pattern/WorkerPool.swift
yim-lee: The same goes for `WorkerPoolSettings` too? I got rid of the `settings` and `selector` properties, and instead store only `logLevel` and `whenAllWorkersTerminated`, which doesn't have any `Worker`.

Sources/DistributedActors/Pattern/WorkerPool.swift
yim-lee: Amended https://github.com/apple/swift-distributed-actors/pull/912/commits/0ed16f138acdfabc6e7891250a385e11b579c4bf

Sources/DistributedActors/Pattern/WorkerPool.swift
yim-lee: Amended https://github.com/apple/swift-distributed-actors/pull/912/commits/0ed16f138acdfabc6e7891250a385e11b579c4bf

Sources/DistributedActors/Pattern/WorkerPool.swift
yim-lee: Amended https://github.com/apple/swift-distributed-actors/pull/912/commits/c987ebe92858fefe6b37f91e1011e9ca065153ba

Sources/DistributedActors/Pattern/WorkerPool.swift
ktoso: Yeah, it makes sense -- it's not a "general" feature as we need lifecycle watch especially.

Sources/DistributedActors/Pattern/WorkerPool.swift
ktoso: I see... I'll have a look if that's maybe compiler bug, thanks!

Sources/DistributedActors/Pattern/WorkerPool.swift
ktoso: Right makes sense... since now storing "refs" (actors) actually forces their extended lifetime, so we can't du stuff we did with refs.

Sources/DistributedActors/Pattern/WorkerPool.swift
ktoso: So `watchTermination` is proactive -- you'll get notified in the moment the actor is deinit/terminated - so we can reap them eagerly. 

Also, for remote references, we need the watch since the refs won't just be deallocated

ktoso:
Looks good, thank you! And we'll follow up with the receptionist mode soon 👍

PR #913, opened 2022-05-29
Periodically prune association tombstones

### LINKED ISSUE #356: Cluster: Prune association tombstones
Associations get tombstoned when we cut connections to them.

We must cleanup the tombstones every `settings.cluster.associationTombstoneTimeToLive` time.
That time is normally set something "very large" so days or many hours. Default now is one day, could be perhaps 12h but let's go with a day for a start I guess (other systems I know use many days, though they're less dynamic than our clusters).

```
    /// Defines when an association shall be completely dropped from the tombstones list.
    /// An association ("unique connection identifier between two nodes") is kept as tombstone when severing a connection between nodes,
    /// in order to avoid accidental re-connections to given node. Once a node has been downed, removed, and disassociated, it MUST NOT be
    // communicated with again. Tombstones are used to ensure this, even if the downed ("zombie") node, attempts to reconnect.
    public var associationTombstoneTimeToLive: TimeAmount = .hours(24) * 1
```

### How to solve

Schedule a periodic (every 10 minutes or so) cleanup tick in ClusterShell and have it skim over the tombstones and other cleanups it may need to do.

drexin:
Resolves #356

### DIFF
--- a/Sources/DistributedActors/ClusterSystemSettings.swift
+++ b/Sources/DistributedActors/ClusterSystemSettings.swift
@@ -115,6 +115,9 @@ public struct ClusterSystemSettings {
     /// communicated with again. Tombstones are used to ensure this, even if the downed ("zombie") node, attempts to reconnect.
     public var associationTombstoneTTL: TimeAmount = .hours(24) * 1
 
+    /// Defines the interval with which the list of associated tombstones is freed from expired tombstones.
+    public var associationTombstoneCleanupInterval: TimeAmount = .minutes(10)
+
     // ==== ------------------------------------------------------------------------------------------------------------
     // MARK: Cluster protocol versioning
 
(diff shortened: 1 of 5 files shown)


drexin:
TODO: Add tests

README.md
```diff
@@ -1,310 +1,14 @@
+# DT Git Hooks
 
-# Swift Distributed Actors
```
drexin, line 2: No idea what happened here, I'll fix it 😅

Sources/DistributedActors/ClusterSystemSettings.swift
```diff
@@ -115,6 +115,9 @@ public struct ClusterSystemSettings {
     /// communicated with again. Tombstones are used to ensure this, even if the downed ("zombie") node, attempts to reconnect.
     public var associationTombstoneTTL: TimeAmount = .hours(24) * 1
 
+    /// Defines the interval with which the list of associated tombstones is freed from expired tombstones.
+    public var associationTombstoneCleanupInterval: TimeAmount = .minutes(10)
```
ktoso, line 119: 10 minutes sounds good 👍

ktoso (approved):

Issue #885, opened 2022-05-17
Confirm async inits already just work™, we fixed this a long time ago


ktoso:
We have a fixme around it from before Kavon did the compiler work;

Make sure there's no issues with async initializers in distributed actors and our system.

Look for: `// FIXME(distributed): this is a bug in the initializers impl, they may call actorReady many times (from async inits) // TODO: probably already solved` and make it crash rather than log warning

PR #920, opened 2022-06-01
Readying actors should not be arbitrarily called many times

ktoso:
https://github.com/apple/swift-distributed-actors/issues/885

We had some bugs about actor ready being called many times by the compiler, and this should have been fixed by now


ktoso:
resolves https://github.com/apple/swift-distributed-actors/issues/885

Issue #281, opened 2019-11-21
Make sure we don't use "public extension" 


ktoso:
> public extensions are very dangerous because the make it so easy to slip in public API by accident.

In general.

PR #921, opened 2022-06-01
[format] Update to swiftformat 0.48.8 and enable extensionacl=on-decls

ktoso:
Resolves https://github.com/apple/swift-distributed-actors/issues/281 which is important so we don't accidentally introduce public API.

The rules are in line with what HTTPClient also does about such: https://github.com/swift-server/async-http-client/commit/e5022468bb41faf0678c9ce9622edf2c1ddf9e64

Sadly it also caused some churn... but better now than later.

### DIFF
--- a/.swiftformat
+++ b/.swiftformat
@@ -1,6 +1,6 @@
 # file options
 
---swiftversion 5.2
+--swiftversion 5.4
 --exclude .build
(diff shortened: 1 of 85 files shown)


ktoso:
@swift-ci please smoke test

Issue #905, opened 2022-05-25
Rephrase AskError as RemoteCallError


ktoso:
There's no "ask" in a DA only world, it is all "remote calls"

This is mostly just renaming.

PR #928, opened 2022-06-01
Rename AskError to RemoteCallError

yim-lee:
Resolves https://github.com/apple/swift-distributed-actors/issues/905

### DIFF
--- a/Sources/DistributedActors/ClusterSystem.swift
+++ b/Sources/DistributedActors/ClusterSystem.swift
@@ -951,7 +951,7 @@ extension ClusterSystem {
         Err: Error
     {
         guard let shell = self._cluster else {
-            throw AskError.systemAlreadyShutDown
+            throw RemoteCallError.systemAlreadyShutDown
         }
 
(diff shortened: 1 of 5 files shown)


Sources/DistributedActors/ClusterSystem.swift
```diff
@@ -1065,8 +1065,11 @@ internal struct LazyStart<Message: ActorMessage> {
     }
 }
 
-enum RemoteCallError: Error {
+enum RemoteCallError: DistributedActorSystemError, Error {
```
yim-lee, line 1068: `RemoteCallError` already exists so I've moved `AskError`'s cases to this. We probably want to keep just one of `clusterAlreadyShutDown` and `systemAlreadyShutDown`?

ktoso: We can stick with `clusterAlreadyShutDown` 👍

ktoso (approved):
Thanks!

PR #934, opened 2022-06-03
Must not touch EL based APIs after shutting down the ELG

### LINKED ISSUE #922: Don't schedule futures on terminated EventLoop
This was added recently: `16:39:10 ERROR: Cannot schedule tasks on an EventLoop that has already shut down. This will be upgraded to a forced crash in future SwiftNIO versions.`

and will become an error soon.

We likely happen to do this because "yolo, doesn't matter" as we're shutting down the actor system, so there's a lot of these happening in tests right now. Avoid doing this by guarding the submissions with the actor system shutdown flag.

ktoso:
Resolves  #922

The newly added repeated task that does tombstone flushing was cancelled AFTER the ELG shutdown, leading to the  `ERROR: Cannot schedule tasks on an EventLoop that has already shut down. This will be upgraded to a forced crash in future SwiftNIO versions.` error explained in Don't schedule futures on terminated EventLoop #922.

We must not touch ELG related APIs after shutting down; A cancel seems to create a future as well, so it was causing issues.

FYI @drexin

### DIFF
--- a/Sources/DistributedActors/ClusterSystem.swift
+++ b/Sources/DistributedActors/ClusterSystem.swift
@@ -454,7 +454,7 @@ public class ClusterSystem: DistributedActorSystem, @unchecked Sendable {
 
             self._associationTombstoneCleanupTask?.cancel()
             self._associationTombstoneCleanupTask = nil
-            
+
             do {
                 try self._eventLoopGroup.syncShutdownGracefully()
                 // self._receptionistRef = self.deadLetters.adapted()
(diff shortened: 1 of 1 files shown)


Sources/DistributedActors/ClusterSystem.swift
```diff
@@ -452,6 +452,9 @@ public class ClusterSystem: DistributedActorSystem, @unchecked Sendable {
             self.systemProvider.stopAll()
             self.dispatcher.shutdown()
 
+            self._associationTombstoneCleanupTask?.cancel()
+            self._associationTombstoneCleanupTask = nil
```
ktoso, line 456: TODO: this write is also not synchronized, would have to be protected somehow.

drexin (approved):
LGTM

Issue #935, opened 2022-06-03
Convert DowningStrategyShell into a distributed actor


ktoso:
It's a pretty simple behavior actor, we should convert it quickly

PR #938, opened 2022-06-03
Convert DowningStrategyShell into a distributed actor

yim-lee:
Resolves https://github.com/apple/swift-distributed-actors/issues/935

### DIFF
--- a/Sources/DistributedActors/Cluster/Downing/DowningStrategy.swift
+++ b/Sources/DistributedActors/Cluster/Downing/DowningStrategy.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift Distributed Actors open source project
 //
-// Copyright (c) 2018-2019 Apple Inc. and the Swift Distributed Actors project authors
+// Copyright (c) 2018-2022 Apple Inc. and the Swift Distributed Actors project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
@@ -12,9 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Distributed
 import Logging
-
-public struct DowningStrategies {}
 
 /// Allows implementing downing strategies, without having to re-implement and reinvent logging and subscription logic.
 ///
@@ -33,7 +32,7 @@ public protocol DowningStrategy {
 public enum DowningStrategyDirective {
     case none
     case markAsDown(Set<Cluster.Member>)
-    case startTimer(key: TimerKey, message: DowningStrategyMessage, delay: TimeAmount)
+    case startTimer(key: TimerKey, member: Cluster.Member, delay: TimeAmount)
     case cancelTimer(key: TimerKey)
 
     static func markAsDown(_ member: Cluster.Member) -> Self {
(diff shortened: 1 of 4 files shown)


Sources/DistributedActors/ClusterSystem.swift
```diff
@@ -331,6 +331,13 @@ public class ClusterSystem: DistributedActorSystem, @unchecked Sendable {
             _ = self._receptionistStore.storeIfNilThenLoad(receptionist)
         }
 
+        // downing strategy (automatic downing)
+        await _Props.$forSpawn.withValue(DowningStrategyShell.props) {
```
yim-lee, line 335: Not sure about move it here

Sources/DistributedActors/Cluster/Downing/DowningStrategy.swift
```diff
@@ -33,88 +32,93 @@ public protocol DowningStrategy {
...
-    case timeout(Cluster.Member)
-}
-
 // ==== ----------------------------------------------------------------------------------------------------------------
 // MARK: Downing Shell
 
-internal struct DowningStrategyShell {
-    typealias Message = DowningStrategyMessage
-    var naming: _ActorNaming = "downingStrategy"
+internal distributed actor DowningStrategyShell: DistributedActor {
```
ktoso, line 46: No need to type out the conformance, it should be implicitly there -- did you see problems with it?

Sources/DistributedActors/Cluster/Downing/DowningStrategy.swift
```diff
@@ -33,88 +32,93 @@ public protocol DowningStrategy {
...
+    typealias ActorSystem = ClusterSystem
 
-    let strategy: DowningStrategy
+    static let path: ActorPath =
+        try! ActorPath([ActorPathSegment("system"), ActorPathSegment("downingStrategy")])
```
ktoso, line 50: I think you can ActorPath._system.appending(...), either is fine just fyi

Sources/DistributedActors/Cluster/Downing/DowningStrategy.swift
```diff
@@ -33,88 +32,93 @@ public protocol DowningStrategy {
...
+            for await event in await system.cluster.events {
+                try self.receiveClusterEvent(event)
             }
         }
     }
 
-    func receiveClusterEvent(_ context: _ActorContext<Message>, event: Cluster.Event) throws {
+    deinit {
+        self.eventsListeningTask?.cancel()
+    }
```
ktoso, line 79: Yup

Sources/DistributedActors/Cluster/Downing/DowningStrategy.swift
```diff
@@ -33,88 +32,93 @@ public protocol DowningStrategy {
...
-            context.timers.startSingle(key: key, message: message, delay: delay)
+        case .startTimer(let key, let member, let delay):
+            self.actorSystem.log.trace("Start timer \(key), member: \(member), delay: \(delay)")
+            self.timers.startSingle(key: key, delay: delay) {
+                self.onTimeout(member: member)
+            }
         case .cancelTimer(let key):
-            context.log.trace("Cancel timer \(key)")
-            context.timers.cancel(for: key)
+            self.actorSystem.log.trace("Cancel timer \(key)")
```
ktoso, line 97: You can create a logger instance inside the actor body like: lazy var logger = Logger(actor: self) this will include its ID in logging then :)

Sources/DistributedActors/ClusterSystem.swift
ktoso: Hmm it probably is fine; previously the reason to be in the shell was to become a child of the cluster actor but we don't have child actors anymore so I think this may be fine. If tests are happy we're happy 😃

ktoso: The entire props dance were able to simplify very soon by the way - the actor system can notice the static props property and pick it up etc

ktoso (approved):
Looks good! I now also realized that it actually is just an actor, not even a distributed one perhaps... but let's keep as is, as we can use our extra logging and other tools with this version

ktoso (approved):
Looks good!

I guess it isn't even a distributed actor but just a normal one huh? But let's keep like this so we can use additional system related APIs like the better logger etc. thanks!

ktoso:
Only one thing about the logger: https://github.com/apple/swift-distributed-actors/pull/938#discussion_r889430590 otherwise ready to go, thank you!

Sources/DistributedActors/Cluster/Downing/DowningStrategy.swift
yim-lee: nope, didn't know.

Sources/DistributedActors/Cluster/Downing/DowningStrategy.swift
yim-lee: amended https://github.com/apple/swift-distributed-actors/pull/938/commits/4dad9a4236115e4bcf82cf9806962b99ef39777d

Sources/DistributedActors/Cluster/Downing/DowningStrategy.swift
ktoso: Ok cool, was worried if some bug 😂

Issue #930, opened 2022-06-02
watchTermination must have async closure version


ktoso:
```
    @discardableResult
    public func watchTermination<Watchee>(
        of watchee: Watchee,
        @_inheritActorContext @_implicitSelfCapture whenTerminated: @escaping @Sendable (ID) -> Void,
        file: String = #file, line: UInt = #line
    ) -> Watchee where Watchee: DistributedActor, Watchee.ActorSystem == ClusterSystem {
```

this must be usable with an async `whenTerminated`

PR #939, opened 2022-06-03
watchTermination should support async 'whenTerminated'

yim-lee:
Resolves https://github.com/apple/swift-distributed-actors/issues/930

### DIFF
--- a/Tests/DistributedActorsTests/LifecycleWatchTests.swift
+++ b/Tests/DistributedActorsTests/LifecycleWatchTests.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift Distributed Actors open source project
 //
-// Copyright (c) 2018-2021 Apple Inc. and the Swift Distributed Actors project authors
+// Copyright (c) 2018-2022 Apple Inc. and the Swift Distributed Actors project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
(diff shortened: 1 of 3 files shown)


Tests/DistributedActorsTests/LifecycleWatchTests.swift
```diff
@@ -71,6 +71,22 @@ distributed actor Juliet: LifecycleWatch, CustomStringConvertible {
         }
     }
 
+    distributed func meetWatchAsyncCallback(
+        _ romeo: Romeo,
+        unwatch doUnwatch: Bool
+    ) async throws {
+        @Sendable
+        func asyncTerminated(_ terminatedIdentity: ClusterSystem.ActorID) async {
+            await self.probe.tell("Received terminated: \(terminatedIdentity)")
```
ktoso, line 80: This await seems not needed? Tell are unidirectional

ktoso (approved):
LGTM!

Tests/DistributedActorsTests/LifecycleWatchTests.swift
yim-lee: Get compilation errors without `await`:

```
/code/Tests/DistributedActorsTests/LifecycleWatchTests.swift:80:13: error: expression is 'async' but is not marked with 'await'
            self.probe.tell("Received terminated: \(terminatedIdentity)")
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
            await 
/code/Tests/DistributedActorsTests/LifecycleWatchTests.swift:80:13: note: property access is 'async'
            self.probe.tell("Received terminated: \(terminatedIdentity)")
```

Removing `@Sendable` gives warnings but compiles (without `await`). Should I keep `@Sendable`?

ktoso: Oh the probe access hmmm

ktoso: Checked out locally: the current code with await triggers a warning about `No 'async' operations occur within 'await' expression` removing the await "just worked" 🤔 

I think we're hitting some difference in toolchains? I'd like to try without the await and see what CI says, that'd be what I'd expect to work hmm

ktoso (approved):

ktoso:
Huh I see, so it seems something changed in sendable semantics on latest nightlies... that's interesting -- merging your take on this API then @yim-lee !

Issue #824, opened 2021-10-12
Implement ActorSingleton for DistributedActor


ktoso:
We need to implement it over the DistributedActor protocol, and remove any ref based APIs.

ktoso:
Completing the implementation depends on https://github.com/apple/swift-distributed-actors/pull/933

ktoso:
I don't think we'll get it within beta-1 but beta-2 easily 👍

PR #940, opened 2022-06-03
Make plugin API async

yim-lee:
Pave way for https://github.com/apple/swift-distributed-actors/issues/824

### DIFF
--- a/Sources/DistributedActors/ClusterSystem.swift
+++ b/Sources/DistributedActors/ClusterSystem.swift
@@ -446,9 +446,7 @@ public class ClusterSystem: DistributedActorSystem, @unchecked Sendable {
             self.cluster.down(member: myselfMember)
         }
 
-        Task {
-            await self.settings.plugins.stopAll(self)
-        }
+        self.settings.plugins.stopAll(self)
 
         queue.async {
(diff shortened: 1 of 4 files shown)


Sources/DistributedActors/ClusterSystem.swift
```diff
@@ -439,7 +439,9 @@ public class ClusterSystem: DistributedActorSystem, @unchecked Sendable {
             self.cluster.down(member: myselfMember)
         }
 
-        self.settings.plugins.stopAll(self)
+        Task {
+            await self.settings.plugins.stopAll(self)
+        }
```
ktoso, line 444: Hmmm we may need to wait for the plugins to stop before we continue shutdown -- so shutdown must become async. We can't spin off a new task I guess as that means concurrent stoppping and shutting down

yim-lee: `stop` is now blocking and non-throwing.

ktoso: Oh I see, then never threw before in reality even hm..

That'll work for now, thanks! May need to become async, we'll see...

ktoso (approved):

PR #941, opened 2022-06-04
Bring back ClusterSystemSettings.enabled

yim-lee:
Partially revert https://github.com/apple/swift-distributed-actors/pull/902 but with `enabled` set to `true` by default instead of `false`.


ktoso:
Thank you! I'm pretty sure the reverting was pretty annoying with some other code having changed since... thanks! 🙏 

We can cleanup the unnecessary `= true` later once we land the other things; having them in helps know what needed them etc

ktoso (approved):

Issue #958, opened 2022-06-08
Remove ActorTimers, and offer task-cancellation and async aware alternative


ktoso:
Depends on https://github.com/apple/swift-async-algorithms/pull/165 Remove the disable of availability and mark methods as available to clock/instant/duration

--- 

The `ActorTimers` cannot easily participate in task cancellation and structured concurrency; we should make use of https://github.com/apple/swift-async-algorithms/blob/main/Sources/AsyncAlgorithms/AsyncAlgorithms.docc/Guides/Timer.md rather than Dispatch -- or rather, end users should be doing this and their TASKS get cancelled, not really actors as they used to in the old behavior world.

This is also the next step of https://github.com/apple/swift-distributed-actors/issues/955 Replace our own TimeAmount with `Swift.Duration` as it simplifies our reliance on those time types.

Probably timers instead, now, thanks to actor re-entrance, tasks and the new Timer inside swift-algorithms is really more of a pattern:

```
Task { 
  for await tick in AsyncTimerSequence.repeating(every: duration) {
    // ... 
    try await receptionist.pushOps(...)
  }
}
```

so this would get cancelled as we decide to cancel the system... we'd have a cancel() on the receptionist...

We don't need to make any new ActorTimers at all I think?

ktoso:
Depends on https://github.com/apple/swift-async-algorithms/pull/165 so we can use the timer async sequence

PR #959, opened 2022-06-08
ActorTimers are not needed in pure DA world #958

ktoso:
Hide the type before removing it completely https://github.com/apple/swift-distributed-actors/issues/958

We'll offer guidance for using async sequences powered by timers instead, see https://github.com/apple/swift-distributed-actors/issues/958


ktoso:
Actually, let's be nicer about those and first deprecate.

Issue #955, opened 2022-06-07
Replace our own TimeAmount with Swift.Duration


ktoso:
Use this everywhere instead: https://developer.apple.com/documentation/swift/duration

### TimeAmount -> Swift.Duration

The purpose of our `TimeAmount` was to not "leak" NIO's TimeAmount, and we should keep true to that.

We should remove the time amount and use `Duration` everywhere.

### Deadline -> ContinuousClock.Instant

I think for our purposes the https://developer.apple.com/documentation/swift/continuousclock will be good enough.

ktoso:
I guess the conversions can be something like:

```
        let (s, `as`) = duration.components // seconds, attoseconds
        let sNanos = s * Int64(1_000_000_000)
        let asNanos = `as` / Int64(1_000_000_000)
        let (totalNanos, overflow) = sNanos.addingReportingOverflow(asNanos)
        
        let nanos: NIO.TimeAmount
        if overflow {
            nanos = .nanoseconds(.max)
        } else {
            nanos = .nanoseconds(totalNanos)
        }
```

we can absolutely just use .max when we get overflow; we never use such large values in practice.

PR #962, opened 2022-06-09
Use Swift's time types

yim-lee:
Resolves https://github.com/apple/swift-distributed-actors/issues/955


Sources/DistributedActors/Time.swift
```diff
@@ -18,116 +18,72 @@ import struct Foundation.Date
...
     /// - parameters:
-    ///     - amount: the amount of hours this `TimeAmount` represents.
-    /// - returns: the `TimeAmount` for the given amount.
-    public static func hours(_ amount: Value) -> TimeAmount {
-        TimeAmount(amount * 1000 * 1000 * 1000 * 60 * 60)
+    ///     - amount: the amount of hours this `Duration` represents.
+    /// - returns: the `Duration` for the given amount.
+    public static func hours(_ amount: Value) -> Duration {
+        .nanoseconds(amount * 1000 * 1000 * 1000 * 60 * 60)
     }
```
ktoso, line 75: This always felt a bit silly to be dropped off from the official type... thanks for adding them

I wonder if we "have to" make them internal...

ktoso: (all the .hours / .seconds thingies i mean)

Sources/DistributedActors/Time.swift
```diff
@@ -209,187 +165,45 @@ extension TimeAmount: CustomStringConvertible, CustomPrettyStringConvertible {
...
-/// - note: `Deadline` should not be used to represent a time interval
-public struct Deadline: Equatable, Hashable {
-    // TODO(swift): To be replaced by types that upcoming Swift versions might gain, as discussed in
-    //              [Pitch] Clock, Instant, Date, and Duration https://forums.swift.org/t/pitch-clock-instant-date-and-duration/52451
-
-    public typealias Value = Int64
-
-    /// The nanoseconds since boot representation of the `Deadline`.
-    public let uptimeNanoseconds: Value
+public typealias Deadline = ContinuousClock.Instant
```
ktoso, line 191: Probably another case where people might be upset at us declaring a public typealias... might have to remove?

Sources/DistributedActors/Time.swift
```diff
@@ -209,187 +165,45 @@ extension TimeAmount: CustomStringConvertible, CustomPrettyStringConvertible {
...
+        .now() - Duration.effectivelyInfinite
     }
 
-    public static let distantPast = Deadline(0)
-    public static let distantFuture = Deadline(Value.max)
-
-    public static func now() -> Deadline {
-        uptimeNanoseconds(Deadline.Value(DispatchTime.now().uptimeNanoseconds))
+    public static var distantFuture: ContinuousClock.Instant {
+        .now() + Duration.effectivelyInfinite
```
ktoso, line 199: are these correct?

Sources/DistributedActors/Time.swift
```diff
@@ -209,187 +165,45 @@ extension TimeAmount: CustomStringConvertible, CustomPrettyStringConvertible {
...
     }
 
-    public static func uptimeNanoseconds(_ nanoseconds: Value) -> Deadline {
-        Deadline(nanoseconds)
+    public static func now() -> ContinuousClock.Instant {
+        .now
     }
```
ktoso, line 204: can probably remove `.now()` in favor of using `.now`

Sources/DistributedActors/Time.swift
```diff
@@ -398,7 +212,7 @@ extension Deadline {
         self.hasTimeLeft(until: .now())
     }
 
-    public func hasTimeLeft(until: Deadline) -> Bool {
+    public func hasTimeLeft(until: ContinuousClock.Instant) -> Bool {
         !self.isBefore(until)
     }
```
ktoso, line 217: I like those... let's keep -- I wonder if people might be upset if we make them public though 🤔

ktoso (approved):
Looks great, only concern about potentially needing to mark extensions on Duration and the typealias `internal` so folks don't get upset we're adding public API to "not our type".

Sources/DistributedActors/Time.swift
yim-lee: Made these internal

Sources/DistributedActors/Time.swift
yim-lee: Hmm I think so (and there are tests)? It used to be `Deadline(Int64.max)`.

Sources/DistributedActors/Time.swift
yim-lee: Made these internal

Sources/DistributedActors/Time.swift
yim-lee: removed

Sources/DistributedActors/Time.swift
ktoso: Strike all that I tried it out and it works out since we cap inside the `.nanoseconds`

Thank you!

Issue #953, opened 2022-06-06
FAILED: DistributedReceptionist test_receptionist_listing_shouldRespondWithRegisteredRefsForKey


ktoso:
Something fishy going on with `listing` API, check the `test_receptionist_listing_shouldRespondWithRegisteredRefsForKey`.

yim-lee:
The test runs successfully if wrapped in `try runAsyncAndBlock` and remove `async` keyword from the test method, but ideally the test should run without these changes.

PR #965, opened 2022-06-09
Re-enable DistributedReceptionistTests.test_receptionist_listing_shouldRespondWithRegisteredRefsForKey

yim-lee:
Resolves https://github.com/apple/swift-distributed-actors/issues/953


yim-lee:
Wrapping test with `try runAsyncAndBlock {` seems to fix it?

yim-lee:
> Wrapping test with try runAsyncAndBlock { seems to fix it?

Maybe not. I added some code to the file that shouldn't impact the test case at all yet it hangs again.

Issue #951, opened 2022-06-06
Make sure cancelling a listing's (outer) Task actually cancels the listing


ktoso:
We don't implement this right now I think but we have to, to support the following pattern:

```swift
distributed actor Boss: LifecycleWatch { 
    var workers: Set<Weak<Worker>> = []
    
    var listingTask: Task<Void, Never>?
    
    func findWorkers() async {
        guard listingTask == nil else {
            actorSystem.log.info("Already looking for workers")
            return
        }

        listingTask = Task {
            for await worker in actorSystem.receptionist.listing(of: .workers) {
                workers.insert(worker)
            }
        }
    }
    
    deinit {
        listingTask?.cancel()
    }
}
```

PR #966, opened 2022-06-09
Make sure to cancel listing when task is cancelled

yim-lee:
Resolves https://github.com/apple/swift-distributed-actors/issues/951

Verified that `DistributedReception.GuestListing.continuation.onTermination` closure gets invoked upon task cancellation, so listing subscription gets cancelled accordingly. Adding a test to make sure no further item is received from the listing.

### DIFF
--- a/Sources/DistributedActors/Timers+Distributed.swift
+++ b/Sources/DistributedActors/Timers+Distributed.swift
@@ -119,8 +119,6 @@ public final class ActorTimers<Act: DistributedActor> where Act.ActorSystem == C
         interval: Duration,
         repeated: Bool
     ) {
-        self.cancel(for: key)
-
         let handle: Cancelable
         if repeated {
             handle = self.dispatchQueue.scheduleAsync(initialDelay: interval, interval: interval) {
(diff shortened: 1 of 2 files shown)


Sources/DistributedActors/Timers+Distributed.swift
```diff
@@ -149,6 +147,9 @@ public final class ActorTimers<Act: DistributedActor> where Act.ActorSystem == C
                 }
 
                 await call()
+
+                // The single timer is done. Remove it so it can be installed again if needed.
+                self.cancel(for: key)
```
ktoso, line 152: This is called from another thread so we need to lock around all updates of     `internal var installedTimers: [TimerKey: DistributedActorTimer] = [:]` now;

or we could turn this into an actor but since this type needs to go away anyway, let's just do the lock to keep it simple.

yim-lee: amended https://github.com/apple/swift-distributed-actors/pull/966/commits/f326d38e60c0540339faa3259ab60eb5bcbfa580

ktoso (approved):
LGTM thank you!

Issue #919, opened 2022-06-01
Re-enable WorkerPoolTests.test_workerPool_registerNewlyStartedActors


yim-lee:
This test fails sometimes because `WorkerPool` doesn't receive all 3 registered workers from `self.actorSystem.receptionist.subscribe` async sequence, and as a result the condition `workers.size() == workerProbes.count` is never true.

```
Test Case 'WorkerPoolTests.test_workerPool_registerNewlyStartedActors' started at 2022-06-01 04:18:00.131
[pprint][4:18:00.1340] [/code/Sources/DistributedActors/Cluster/ClusterShell.swift:271][thread:139629990770432]: Started ClusterShell
2022-06-01T04:18:00+0000 warning /system/receptionist : actor/id=/system/receptionist actor/path=/system/receptionist distributed receptionist: register(DistributedActorsTests.(unknown context at $55729e4a7dfc).Greeter, with: DistributedReception.Key<DistributedActorsTests.(unknown context at $55729e4a7dfc).Greeter>(id: request-workers)
2022-06-01T04:18:00+0000 warning /system/receptionist : actor/id=/system/receptionist actor/path=/system/receptionist distributed receptionist: register(DistributedActorsTests.(unknown context at $55729e4a7dfc).Greeter, with: DistributedReception.Key<DistributedActorsTests.(unknown context at $55729e4a7dfc).Greeter>(id: request-workers)
2022-06-01T04:18:00+0000 warning /system/receptionist : actor/id=/system/receptionist actor/path=/system/receptionist distributed receptionist: register(DistributedActorsTests.(unknown context at $55729e4a7dfc).Greeter, with: DistributedReception.Key<DistributedActorsTests.(unknown context at $55729e4a7dfc).Greeter>(id: request-workers)
/code/Tests/DistributedActorsTests/Pattern/WorkerPoolTests.swift:55: error: WorkerPoolTests.test_workerPool_registerNewlyStartedActors : Asynchronous wait failed - Exceeded timeout of 3.0 seconds, with unfulfilled expectations: all workers available
/code/Tests/DistributedActorsTests/Pattern/WorkerPoolTests.swift:66: error: WorkerPoolTests.test_workerPool_registerNewlyStartedActors : failed - Did not receive expected [work:2 at /user/Greeter-n]:String within [5s], error: noMessagesInQueue
<EXPR>:0: error: WorkerPoolTests.test_workerPool_registerNewlyStartedActors : threw error "
[...]

PR #967, opened 2022-06-10
Receptionist should flush registrations in order by version

yim-lee:
Resolves https://github.com/apple/swift-distributed-actors/issues/919


ktoso (approved):
Ah I see, since we're powering a stream with dedup now... and the behavior implementation just dumped the entire listing always but now we need to dedup here... 

Maybe we can get this at lower cost sometime, but this looks good, thank you!

Issue #948, opened 2022-06-06
Add an API to "await joining the cluster"


ktoso:
It'd need to subscribe to events and return as soon as we get an `.up` about our node.

ktoso:
In general we likely can now offer a facility to await a status of any `node`, that'll be quite useful.

Previously we only had this in tests -- see the blocking implementations in `assertMemberStatus` in the TestKit. We should instead make those `async/await` and offer them in the main cluster module.

PR #969, opened 2022-06-10
Make ClusterControl.membershipSnapshot async property

yim-lee:
Resolves https://github.com/apple/swift-distributed-actors/issues/949

PR #972, opened 2022-06-10
Add API to 'await joining cluster'

yim-lee:
Resolves https://github.com/apple/swift-distributed-actors/issues/948

~~Depends on https://github.com/apple/swift-distributed-actors/pull/969. Please look at https://github.com/apple/swift-distributed-actors/commit/8f46d0472dc92fa0f68289814d46785ea35d0af8 only.~~

### DIFF
--- a/Sources/DistributedActors/Cluster/Cluster+Membership.swift
+++ b/Sources/DistributedActors/Cluster/Cluster+Membership.swift
@@ -663,11 +663,36 @@ extension MembershipDiff: CustomDebugStringConvertible {
 // MARK: Errors
 
 extension Cluster {
-    public enum MembershipError: Error {
+    public enum MembershipError: Error, CustomStringConvertible {
         case nonMemberLeaderSelected(Cluster.Membership, wannabeLeader: Cluster.Member)
-        case notFound(UniqueNode)
+        case notFound(UniqueNode, in: Cluster.Membership)
         case atLeastStatusRequirementNotMet(expectedAtLeast: Cluster.MemberStatus, found: Cluster.Member)
         case statusRequirementNotMet(expected: Cluster.MemberStatus, found: Cluster.Member)
         case awaitStatusTimedOut(Duration, Error?)
+
+        public var description: String {
+            switch self {
+            case .nonMemberLeaderSelected(let membership, let wannabeLeader):
+                return "[\(wannabeLeader)] selected leader but is not a member [\(membership)]"
+            case .notFound(let node, let membership):
+                return "[\(node)] is not a member [\(membership)]"
+            case .atLeastStatusRequirementNotMet(let expectedAtLeastStatus, let foundMember):
+                return "Expected \(reflecting: foundMember.uniqueNode) to be seen as at-least [\(expectedAtLeastStatus)] but was [\(foundMember.status)]"
+            case .statusRequirementNotMet(let expectedStatus, let foundMember):
+                return "Expected \(reflecting: foundMember.uniqueNode) to be seen as [\(expectedStatus)] but was [\(foundMember.status)]"
+            case .awaitStatusTimedOut(let duration, let lastError):
+                let lastErrorMessage: String
+                if let error = lastError {
+                    lastErrorMessage = "Last error: \(error)"
+                } else {
+                    lastErrorMessage = "Last error: <none>"
+                }
+
+                return """
+                No result within \(duration.prettyDescription).
+                \(lastErrorMessage)
+                """
+            }
+        }
     }
 }
--- a/Sources/DistributedActors/Cluster/Cluster+Membership.swift
+++ b/Sources/DistributedActors/Cluster/Cluster+Membership.swift
@@ -666,8 +666,8 @@ extension Cluster {
     public enum MembershipError: Error {
         case nonMemberLeaderSelected(Cluster.Membership, wannabeLeader: Cluster.Member)
         case notFound(UniqueNode)
-        case atLeastStatusRequirementNotMet(expectedAtLeast: Cluster.MemberStatus, actual: Cluster.MemberStatus)
-        case statusRequirementNotMet(expected: Cluster.MemberStatus, actual: Cluster.MemberStatus)
+        case atLeastStatusRequirementNotMet(expectedAtLeast: Cluster.MemberStatus, found: Cluster.Member)
+        case statusRequirementNotMet(expected: Cluster.MemberStatus, found: Cluster.Member)
         case awaitStatusTimedOut(Duration, Error?)
     }
 }
--- a/Sources/DistributedActors/Cluster/Cluster+Membership.swift
+++ b/Sources/DistributedActors/Cluster/Cluster+Membership.swift
@@ -665,5 +665,9 @@ extension MembershipDiff: CustomDebugStringConvertible {
 extension Cluster {
     public enum MembershipError: Error {
         case nonMemberLeaderSelected(Cluster.Membership, wannabeLeader: Cluster.Member)
+        case notFound(UniqueNode)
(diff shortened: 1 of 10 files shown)


Sources/DistributedActors/Cluster/ClusterControl.swift
```diff
@@ -145,4 +155,77 @@ public struct ClusterControl {
...
+                return
+            } catch {
+                lastError = error
+                try await Task.sleep(nanoseconds: UInt64(interval.nanoseconds))
+            }
+        }
+
+        throw Cluster.MembershipError.awaitStatusTimedOut(within, lastError)
+    }
 }
```
yim-lee, line 231: @ktoso Are these APIs what you have in mind?

Sources/DistributedActors/Cluster/ClusterControl.swift
```diff
@@ -155,4 +155,77 @@ public struct ClusterControl {
     public func down(member: Cluster.Member) {
         self.ref.tell(.command(.downCommandMember(member)))
     }
+
+    public func joined(node: UniqueNode, within: Duration) async throws {
+        try await self.awaitMemberStatus(is: .up, within: within, node: node)
+    }
```
ktoso, line 161: That's a nice one; past tense works well here since it is `joining -> up` so makes sense; It'll be a common API to use I suspect, good to have a special version for it.

Can we perhaps also return the `Member` for the joined node?

ktoso: PS: Good to always force passing a duration, agreed 👍

Sources/DistributedActors/Cluster/ClusterControl.swift
```diff
@@ -155,4 +155,77 @@ public struct ClusterControl {
...
+
+    public func ensureNodes(_ status: Cluster.MemberStatus, within: Duration, nodes: [UniqueNode]) async throws {
```
ktoso, line 163: Good API but some rewordings perhaps...? Since we're pulling it out from a testing API to a full API maybe we can make it a bit nicer.

```
func waitFor(nodes: Set<UniqueNode>, _ status: Cluster.MemberStatus, within: Duration) async throws
func waitFor(nodes: Set<UniqueNode>, atLeast status: Cluster.MemberStatus, within: Duration) async throws
```

we need the "exact" and "at least" versions I think. At least means "the nodes are at least `.joining`" etc.

When the exact API sees at least one node that is already PAST the expected/requested status we can throw; but the other one we can still allow and succeed, say, a node was up but became down already and with `atLeast: .up` that'd be okey.

Sources/DistributedActors/Cluster/ClusterControl.swift
```diff
@@ -155,4 +155,77 @@ public struct ClusterControl {
...
+        try await withThrowingTaskGroup(of: Void.self) { group in
+            for node in nodes {
+                group.addTask {
+                    try await self.awaitMemberStatus(is: status, within: within, node: node)
+                }
+                try await group.next() // this is required to propagate error and cancel all tasks
```
ktoso, line 169: not required for cancelling; but yes required for the error propagation -- slight fixup in comment here only

ktoso: Slighly change this API perhaps, so the group actually waits concurrently:

```swift
            for node in nodes {
                 group.addTask {
                     try await self.awaitMemberStatus(is: status, within: within, node: node)
                 }
            }

// loop explicitly to propagagte any throws that might have happened 
            for try await status in group { nothing }
```

that try await on the group does what we need here

Sources/DistributedActors/Cluster/ClusterControl.swift
ktoso: Ah sorry I see you did the atLeast as well already 👍  (was reading top down)

Sources/DistributedActors/Cluster/ClusterControl.swift
```diff
@@ -155,4 +155,77 @@ public struct ClusterControl {
...
+            for node in nodes {
+                group.addTask {
+                    try await self.awaitMemberStatus(atLeast: atLeastStatus, within: within, node: node)
+                }
+                try await group.next() // this is required to propagate error and cancel all tasks
+            }
+        }
+    }
+
+    public func awaitMemberStatus(atLeast atLeastStatus: Cluster.MemberStatus, within: Duration, node: UniqueNode) async throws {
```
ktoso, line 185: Naming seems inconsistent (maybe it was inconsistent in the testkit, sorry if so).

Let's try these shapes maybe:
- `try await system.cluster.waitFor(node, .up, within: ...)`
    - `try await system.cluster.waitFor(node, atLeast: .up, within: ...)`
- `try await system.cluster.waitFor([n1, n2, n2], .up, within: ...)`
    - `try await system.cluster.waitFor([n1, n2, n2], atLeast: .up, within: ...)`

Sources/DistributedActors/Cluster/ClusterControl.swift
ktoso: Yes those look mostly good, commented a little bit on the shapes; I guess the testkit ones may have been a bit inconsistent?

Sources/DistributedActors/Cluster/ClusterControl.swift
yim-lee: oops, good catch

Sources/DistributedActors/Cluster/ClusterControl.swift
```diff
@@ -155,4 +155,116 @@ public struct ClusterControl {
...
+    ///   - status: The expected member status.
+    ///   - within: Duration to wait for.
+    ///
+    /// - Returns `Cluster.Member` for the joined node.
+    public func waitFor(_ node: UniqueNode, _ status: Cluster.MemberStatus, within: Duration) async throws -> Cluster.Member {
+        try await self.waitForMembershipEventually(within: within) { membership in
+            guard let foundMember = membership.uniqueMember(node) else {
+                throw Cluster.MembershipError.notFound(node, in: membership)
+            }
+
```
ktoso, line 219: Oh sorry to bother you but I thought of another case...

If a node was removed BEFORE, we'll have a tombstone for it inside the `    private var _associationTombstones: [UniqueNode: Association.Tombstone]` protected by `_associationsLock`. If it is in there, we should return that the member was `.removed`

You can "make up" a member for a removed node by creating a `Member(` from the unique node and put `.unreachable` and `.removed` in there 👍 

This could be in a separate PR  👍

Sources/DistributedActors/Cluster/Cluster+Membership.swift
```diff
@@ -663,7 +663,36 @@ extension MembershipDiff: CustomDebugStringConvertible {
...
+                } else {
+                    lastErrorMessage = "Last error: <none>"
+                }
+
+                return """
+                No result within \(duration.prettyDescription).
+                \(lastErrorMessage)
+                """
+            }
+        }
```
ktoso, line 696: Looks good but Perhaps must be wrapped in `"MembershipError(\(self), details: <the nice strings>")`? Otherwise might be weird not seeing what type the error was?

ktoso (approved):
LGTM!

Minor potential addition to handle removed nodes better

Sources/DistributedActors/Cluster/ClusterControl.swift
yim-lee: amended https://github.com/apple/swift-distributed-actors/pull/972/commits/5178fa44108777eb2c921ac5fd9279f048ee6bab

ktoso:
Will amend the wording, look at the failure and merge, thank you!  :)

Sources/DistributedActors/Cluster/ClusterControl.swift
ktoso: Looking good!

Issue #970, opened 2022-06-10
FAILED: InterceptorTests.test_interceptor_shouldInterceptSignals


yim-lee:
```
11:02:19 Test Case 'InterceptorTests.test_interceptor_shouldInterceptSignals' started at 2022-06-10 18:02:19.205
11:02:19 <EXPR>:0: error: InterceptorTests.test_interceptor_shouldInterceptSignals : threw error "
11:02:19         try p.expectNoMessage(for: .milliseconds(500))
11:02:19              ^~~~~~~~~~~~~~~~
11:02:19 error: Received unexpected message [_ChildTerminated(/user/theWallsHaveEarsForTermination/stopperTwo)]:_ChildTerminated. Did not expect to receive any messages for [500ms]."
11:02:19 ------------------------------------- ClusterSystem(InterceptorTests) ------------------------------------------------
11:02:19 Captured log [InterceptorTests][2022-06-10 18:2:19.2100] [ClusterSystem.swift:909] [warning] Assign identity
11:02:19 // metadata:
11:02:19 // "actor/id": /system/receptionist
11:02:19 // "actor/id/uniqueNode": sact://InterceptorTests@127.0.0.1:9001
11:02:19 // "actor/type": OpLogDistributedReceptionist
11:02:19 Captured log [InterceptorTests][2022-06-10 18:2:19.2100] [ClusterSystem.swift:922] [trace] Actor ready
11:02:19 // metadata:
11:02:19 // "actor/id": /system/receptionist
11:02:19 // "actor/type": OpLogDistributedReceptionist
11:02:19 Captured log [InterceptorTests][2022-06-10 18:2:19.2110] [ClusterSystem.swift:395] [info] ClusterSystem [InterceptorTests] initialized, listening on: sact://InterceptorTests@127.0.0.1:9001
11:02:19 Captured log [InterceptorTests][2022-06-10 18:2:19.2110] [_OperationLogClusterReceptionistBehavior.swift:98][/system/receptionist-ref] [debug] Initialized receptionist
11:02:19 Captured log [InterceptorTests][2022-06-10 18:2:19.2110] [ClusterEventStream.swift:80][/system/clusterEvents] [trace] Successfully added async subscriber [ObjectIdentifier(0x00007f3e846bce30)]
[...]

PR #976, opened 2022-06-17
!context store lifecycle in DistributedActorContext; internal only

### LINKED ISSUE #926: Move storage of LifecycleWatch into DistributedActorContext
`DistributedActorContext` is the new `ActorContext` where we can store state for distributed actors.

It is internal only and stored inside the actor's identity.

The reason this matters is because today we're hitting a horrible hack to store all lifecycles in a hashmap in the actor system, but they should be in the identity instead, as they are "per actor".

State is to be initialized during actorRady

ktoso:
Resolves #926

By moving state associated with concrete actors, we decrease lock contention and allow each actor to store the additional state used for lifecycle API.

We do not expose the distributed actor context as user API, and it will only be ever accessed by the actor system at actor init and actor deinit and the actor itself via the LifecycleWatch protocol


ktoso:
(silly bugs remain, but will fix)

ktoso:

ktoso:

ktoso:

ktoso:
signal intercepting likes to flaky fail, but we're removing this anyway 🤔

ktoso:

ktoso:
Flaky failure #970

Issue #973, opened 2022-06-14
Convert SWIMActorShell into distributed actor


yim-lee:

ktoso:
Let's land this in beta.3 :)

PR #977, opened 2022-06-18
Convert SWIMActorShell to distributed actor

yim-lee:
Resolves https://github.com/apple/swift-distributed-actors/issues/973

⚠️ ~~Haven't worked on tests yet~~ ⚠️


Sources/DistributedActors/Cluster/SWIM/SWIMPeer+Actor.swift
```diff
@@ -18,16 +18,116 @@ import struct Dispatch.DispatchTime
...
+}
+
+extension SWIMActorShell: SWIMPingOriginPeer {
+    nonisolated func ack(
+        acknowledging sequenceNumber: SWIM.SequenceNumber,
+        target: SWIMPeer,
+        incarnation: SWIM.Incarnation,
+        payload: SWIM.GossipPayload
+    ) {
+        fatalError("This implementation doesn't send ack directly")
```
yim-lee, line 105: Some SWIM implementations need explicit `ack`/`nack` response to trigger callbacks. This implementation doesn't need it because response is returned as the result of `ping`/`pingRequest`.

Sources/DistributedActors/Cluster/SWIM/Protobuf/SWIM+Serialization.swift
```diff
@@ -165,18 +165,18 @@ extension SWIM.Member: _ProtobufRepresentable {
...
+        proto.id = try actorPeer.id.toProto(context: context)
         proto.status = try self.status.toProto(context: context)
         proto.protocolPeriod = self.protocolPeriod
         return proto
     }
 
     public init(fromProto proto: _ProtoSWIMMember, context: Serialization.Context) throws {
         let id = try ActorID(fromProto: proto.id, context: context)
-        let peer = context._resolveActorRef(SWIM.Message.self, identifiedBy: id)
+        let peer = try SWIM.Actor.resolve(id: id, using: context.system)
```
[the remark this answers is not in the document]
ktoso: Yes, this is correct `(SomeActorType).resolve` is the way to do these now.

Nitpick on the name `Actor` as type name is a bit confusing I guess since there is an `protocol Actor` in stdlib

Sources/DistributedActors/Cluster/SWIM/SWIM+Messages.swift
```diff
@@ -15,6 +15,7 @@
 import ClusterMembership
 import SWIM
 
+/*
```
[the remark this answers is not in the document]
ktoso: Yes, I think so -- since the messages are now distributed funcs

Sources/DistributedActors/Cluster/SWIM/SWIMActorShell.swift
```diff
@@ -541,16 +1049,25 @@ internal struct SWIMActorShell {
...
     static let name: String = "swim"
     static let naming: _ActorNaming = .unique(SWIMActorShell.name)
 
     static var props: _Props {
-        _Props
-            ._wellKnown
-            .metrics(group: "swim.shell", measure: [.serialization, .deserialization])
+        var ps = _Props()
+        ps._knownActorName = ActorPath._swim.name
+//        ps._systemActor = true
```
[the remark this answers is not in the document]
ktoso: I guess not in the "new world" since system/not system really was about "where in the tree" they live, and since we don't have the tree in DA world, it doesn't matter as much. Ok to not have it as system.

Sources/DistributedActors/Cluster/SWIM/SWIMPeer+Actor.swift
ktoso: That's good with me 👍

Sources/DistributedActors/Cluster/ClusterShell.swift
```diff
@@ -119,7 +119,11 @@ internal class ClusterShell {
         }
 
         // 2) Ensure the failure detector knows about this node
-        self._swimRef?.tell(.local(SWIM.LocalMessage.monitor(associated.handshake.remoteNode)))
+        Task {
+            await self._swimShell?.whenLocal { swim in
+                swim.monitor(node: associated.handshake.remoteNode)
+            }
```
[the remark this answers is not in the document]
ktoso: Yeah that's right though implementation is lacking and you'll need to do ```whenLocal { in __secretlyKnownToBeLocal in` (that name matters... it's  a hack until we implement https://github.com/apple/swift/pull/42098 which is now my top compiler priority as we implemented all the other lagging things

Sources/DistributedActors/Cluster/ClusterShell.swift
```diff
@@ -661,7 +682,11 @@ extension ClusterShell {
 
     func tryConfirmDeadToSWIM(_ context: _ActorContext<Message>, _ state: ClusterShellState, change: Cluster.MembershipChange) {
         if change.status.isAtLeast(.down) {
-            self._swimRef?.tell(.local(SWIM.LocalMessage.confirmDead(change.member.uniqueNode)))
+            Task {
+                await self._swimShell?.whenLocal { swim in
```
ktoso, line 686:
```suggestion
                await self._swimShell?.whenLocal { __secretlyKnownToBeLocal in // TODO(distributed): rename once https://github.com/apple/swift/pull/42098 is implemented
```

Sources/DistributedActors/Cluster/ClusterShell.swift
```diff
@@ -1269,7 +1294,11 @@ extension ClusterShell {
         }
 
         // whenever we down a node we must ensure to confirm it to swim, so it won't keep monitoring it forever needlessly
-        self._swimRef?.tell(.local(SWIM.LocalMessage.confirmDead(memberToDown.uniqueNode)))
+        Task {
+            await self._swimShell?.whenLocal { swim in
```
ktoso, line 1298:
```suggestion
            await self._swimShell?.whenLocal { __secretlyKnownToBeLocal in // TODO(distributed): rename once https://github.com/apple/swift/pull/42098 is implemented
```

Sources/DistributedActors/Cluster/SWIM/ClusterMembership+Converters.swift
```diff
@@ -27,9 +27,13 @@ extension ClusterMembership.Node {
...
-
-    func swimRef(_ context: _ActorContext<SWIM.Ref.Message>) -> SWIM.PeerRef {
-        context.system._resolve(context: .init(id: ._swim(on: self.asUniqueNode!), system: context.system)) // TODO: the ! is not so nice
+//
+//    func swimRef(_ context: _ActorContext<SWIM.Ref.Message>) -> SWIM.PeerRef {
+//        context.system._resolve(context: .init(id: ._swim(on: self.asUniqueNode!), system: context.system)) // TODO: the ! is not so nice
+//    }
+//    
+    func swimActor(_ system: ClusterSystem) -> SWIM.Actor {
+        try! SWIM.Actor.resolve(id: ._swim(on: self.asUniqueNode!), using: system) // TODO: the ! is not so nice
```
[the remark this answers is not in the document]
ktoso: Yes that is right; In DA world the `SomeType.resolve` is what used to be the cluster's internal resolve functions 👍

Sources/DistributedActors/Cluster/SWIM/SWIMActorShell.swift
```diff
@@ -15,9 +15,517 @@
...
+    var metrics: SWIM.Metrics {
+        self.swim.metrics
+    }
+
+    init(settings: SWIM.Settings, clusterRef: ClusterShell.Ref, system: ActorSystem) async {
+        self._settings = settings
+        self.clusterRef = clusterRef
+        self.actorSystem = system
+        
+        self.onStart()
```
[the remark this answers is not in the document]
ktoso: With the async there is no warning right? That's what I'd expect and that's fine -- yeah ok to have the init async; it is unfortunate but those are the actor semantics for inits...

Sources/DistributedActors/Cluster/SWIM/SWIMActorShell.swift
```diff
@@ -15,9 +15,517 @@
...
+            "swim/timeout": "\(timeout)",
+        ]))
+
+        let pingSentAt = DispatchTime.now()
+        self.metrics.shell.messageOutboundCount.increment()
+        
+        do {
+            let response = try await target.ping(payload: payload, from: self, timeout: timeout, sequenceNumber: sequenceNumber)
+            self.metrics.shell.pingResponseTime.recordInterval(since: pingSentAt)
+            return self.handlePingResponse(
```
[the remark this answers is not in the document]
ktoso: That's very cool actually, great example how request/response focused API of DA makes such patterns simpler :)

Sources/DistributedActors/Cluster/ClusterShell.swift
```diff
@@ -149,9 +153,13 @@ internal class ClusterShell {
...
         traceLog_Remote(system.cluster.uniqueNode, "Finish terminate association [\(remoteNode)]: Notifying SWIM, .confirmDead")
-        if let swim = self._swimRef {
+        if let swim = self._swimShell {
             system.log.warning("Confirm .dead to underlying SWIM, node: \(reflecting: remoteNode)")
-            swim.tell(.local(SWIM.LocalMessage.confirmDead(remoteNode)))
+            Task {
+                await swim.whenLocal { swim in
+                    swim.confirmDead(node: remoteNode)
+                }
+            }
```
ktoso, line 162: To make this look a bit less horrible you could

```
nonisolated func confirmDead(node ...) {
  Task { await ... // same as here }
}
```

onto the swim shell, so the call site will look nice: just `swim.confirmDead(...)`

Sources/DistributedActors/Cluster/ClusterShell.swift
```diff
@@ -396,8 +412,13 @@ extension ClusterShell {
...
             // 1) failure detector:
-            let swimBehavior = SWIMActorShell.behavior(settings: self.settings.swim, clusterRef: context.myself)
-            self._swimRef = try context._spawn(SWIMActorShell.naming, props: SWIMActorShell.props, swimBehavior)
+//            let swimBehavior = SWIMActorShell.behavior(settings: self.settings.swim, clusterRef: context.myself)
+//            self._swimRef = try context._spawn(SWIMActorShell.naming, props: SWIMActorShell.props, swimBehavior)
+            Task {
+                await _Props.$forSpawn.withValue(SWIMActorShell.props) {
+                    self._swim = await SWIMActorShell(settings: self.settings.swim, clusterRef: context.myself, system: context.system)
+                }
+            }
```
ktoso, line 421: Can we try to make the cluster shell init async as well, this way we don't have to do the lock dance?

If definite initialization wil be preventing us we could get away by marking the `swim` var as `var swim: SWIM.Shell!` so we would be allowed to `await` and THEN initialize it, I think then?

Sources/DistributedActors/Cluster/SWIM/SWIMActorShell.swift
```diff
@@ -15,9 +15,517 @@
...
+                switch directive {
+                case .gossipProcessed(let gossipDirective):
+                    self.handleGossipPayloadProcessedDirective(gossipDirective)
+
+                case .sendAck(_, let pingedTarget, let incarnation, let payload, let sequenceNumber):
+                    continuation.resume(returning: .ack(target: pingedTarget, incarnation: incarnation, payload: payload, sequenceNumber: sequenceNumber))
+                }
+            }
+        }
+    }
```
ktoso, line 420: The continuation is here only because forEach isn't async right?

We can/should use a for loop though then we'd be able to avoid the continuation and just return 👍

Sources/DistributedActors/Cluster/SWIM/SWIMActorShell.swift
```diff
@@ -15,9 +15,517 @@
...
+                            pingRequestOrigin: pingRequestOrigin,
+                            pingRequestSequenceNumber: pingRequestSequenceNumber,
+                            timeout: timeout,
+                            sequenceNumber: pingSequenceNumber
+                        )
+                        continuation.resume(returning: response)
+                    }
+                }
+            }
+        }
```
ktoso, line 459: Turns out using the swim instance is hard to use with async await heh I guess :-(

ktoso:
Hmmm async await makes things quite tough here :-\
Thank you for working on this, it really helps!

Sources/DistributedActors/Cluster/SWIM/SWIMActorShell.swift
```diff
@@ -15,9 +15,517 @@
...
+            ]))
+
+            self.swim.onPing(
+                pingOrigin: origin,
+                payload: payload,
+                sequenceNumber: sequenceNumber
+            ).forEach { directive in
+                switch directive {
+                case .gossipProcessed(let gossipDirective):
+                    self.handleGossipPayloadProcessedDirective(gossipDirective)
```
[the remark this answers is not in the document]
ktoso: Oh wait...! 

Is it actually that these always return the `.sendAck` directive so we're always guaranteed to continue it? Indeed:

```

    public func onPing(pingOrigin: SWIMPingOriginPeer, payload: SWIM.GossipPayload, sequenceNumber: SWIM.SequenceNumber) -> [PingDirective] {
        var directives: [PingDirective]

        // 1) Process gossip
        directives = self.onGossipPayload(payload).map { g in
            .gossipProcessed(g)
        }

        // 2) Prepare reply
        directives.append(.sendAck(
            to: pingOrigin,
            pingedTarget: self.peer,
            incarnation: self.incarnation,
            payload: self.makeGossipPayload(to: pingOrigin),
            acknowledging: sequenceNumber
        ))

        return directives
    }
```

So... this is indeed correct, but by looking at the code it looks as if it wasn't...

Perhaps let's make it a for loop anyway, and `assert()` that the returned directives contain the `.sendPing` directive and call it ok-enough for now?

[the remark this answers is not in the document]
ktoso: Maybe...

```swift
for directive in swim.onPing(...) {
  switch directive { 
  case .gossipProcessed(...):
  case .sendPing(...):
  return .ack(target: pingedTarget, incarnation: incarnation, payload: payload, sequenceNumber: sequenceNumber)
}
preconditionFailure("should have sent ack")
```

could work?

Sources/DistributedActors/Cluster/SWIM/SWIMActorShell.swift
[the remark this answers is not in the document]
yim-lee: > With the async there is no warning right?

Right, no warning with `async`.

> we could keep this self.onStart out of the init and ask the caller to kick it off... if that makes the init dance easier?

Not much easier I guess? The caller would have to remember to make the call and do the local method dance.

Sources/DistributedActors/Cluster/ClusterShell.swift
yim-lee: It doesn't allow me to:

```
/code/Sources/DistributedActors/Cluster/SWIM/SWIMActorShell.swift:475:34: error: distributed actor-isolated property 'swim' can not be accessed from a non-isolated context
            let directive = self.swim.confirmDead(peer: node.asSWIMNode.swimActor(self.actorSystem))
```

Sources/DistributedActors/Cluster/ClusterShell.swift
```diff
@@ -277,33 +284,17 @@ internal class ClusterShell {
     }
 
     /// Actually starts the shell which kicks off binding to a port, and all further cluster work
-    internal func lazyStart(system: ClusterSystem, clusterEvents: EventStream<Cluster.Event>) throws -> LazyStart<Message> {
```
yim-lee, line 280: This doesn't look like it gets used, so removing it.

Sources/DistributedActors/Cluster/ClusterShell.swift
yim-lee: How about https://github.com/apple/swift-distributed-actors/pull/977/commits/d862fb20b5112b60411b02205115e5ad36ea6ab8?

Sources/DistributedActors/Cluster/SWIM/SWIMActorShell.swift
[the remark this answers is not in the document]
yim-lee: I also debated between getting response through APIs in SWIM protocol (the current impl) vs. calling distributed APIs/funcs directly. In the latter case though, we would not be calling the protocol at all and I am not sure we would like that.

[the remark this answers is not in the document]
yim-lee: > Is it actually that these always return the .sendAck directive so we're always guaranteed to continue it? 

Yeah, the distributed func `ping` should always return `ack` and `pingRequest` should always return either `ack` or `nack`. If `ping` fails (e.g., timeout), the shell will trigger `pingRequest` call(s).

Sources/DistributedActors/Cluster/SWIM/SWIMActorShell.swift
```diff
@@ -15,9 +15,517 @@
...
+
+            self.swim.onPingRequest(
+                target: target,
+                pingRequestOrigin: pingRequestOrigin,
+                payload: payload,
+                sequenceNumber: pingRequestSequenceNumber
+            ).forEach { directive in
+                switch directive {
+                case .gossipProcessed(let gossipDirective):
+                    self.handleGossipPayloadProcessedDirective(gossipDirective)
```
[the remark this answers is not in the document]
yim-lee: removed continuation: https://github.com/apple/swift-distributed-actors/pull/977/commits/e238478d23e6eb83298e6a721b818783c2d99192

Sources/DistributedActors/Cluster/SWIM/SWIMActorShell.swift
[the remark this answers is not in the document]
yim-lee: amended https://github.com/apple/swift-distributed-actors/pull/977/commits/e238478d23e6eb83298e6a721b818783c2d99192

Sources/DistributedActors/Cluster/SWIM/SWIMActorShell.swift
yim-lee: removed continuation https://github.com/apple/swift-distributed-actors/pull/977/commits/e238478d23e6eb83298e6a721b818783c2d99192

Sources/DistributedActors/Cluster/ClusterShell.swift
ktoso: I don't understand that one -- `self.swim` is on the shell right?

Sources/DistributedActors/Cluster/ClusterShell.swift
ktoso: nice :)

Sources/DistributedActors/Cluster/ClusterShell.swift
yim-lee: Sorry, I didn't do completely as you prescribed. Got it to work now.

Sources/DistributedActors/ClusterSystem.swift
```diff
@@ -330,7 +330,7 @@ public class ClusterSystem: DistributedActorSystem, @unchecked Sendable {
                 systemStream: true,
                 customBehavior: ClusterEventStream.Shell.behavior
             )
-            let clusterRef = try! cluster.start(system: self, clusterEvents: clusterEvents) // only spawns when cluster is initialized
+            let clusterRef = try! await cluster.start(system: self, clusterEvents: clusterEvents) // only spawns when cluster is initialized
```
yim-lee, line 333: The changes in this PR have disrupted the shutdown flow, because the error has come back:

```
ERROR: Cannot schedule tasks on an EventLoop that has already shut down. This will be upgraded to a forced crash in future SwiftNIO versions.
```

Sources/DistributedActors/Cluster/ClusterShell.swift
ktoso: Uff okey! Was worried we had some isolation issue -- glad it works :)

Sources/DistributedActors/ClusterSystem.swift
ktoso: Okey :-/

I think for this PR feel free to ignore those and I'll see about it separately again 👍 
We need some more solid way to prevent this...

Sources/DistributedActors/Cluster/SWIM/Protobuf/SWIM+Serialization.swift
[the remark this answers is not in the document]
ktoso: It looks better than `SWIM.Actor` I think, it is a bit "meh" since a lot of types end up being called "Actor" if we go down that path... We'll find some better name -- maybe just SWIMActor or SWIMActorShell eventually?>?

In this PR just wanted to avoid the Actor` typealias

Sources/DistributedActors/Cluster/SWIM/SWIMActorShell.swift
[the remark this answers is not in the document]
ktoso: The amended version looks better, thanks!

It is a bit weird with those directives now but it actually reads okey... and we codified the assumptions between the SWIM and our impl, looks good to me 👍

Sources/DistributedActors/ClusterSystem.swift
yim-lee: This helps: https://github.com/apple/swift-distributed-actors/pull/977/commits/810cf91841c9b0dc38e6986a2a2336f959118816

Tests/DistributedActorsTests/Cluster/SWIM/Protobuf/SWIM+SerializationTests.swift
```diff
@@ -18,32 +18,34 @@ import SWIM
...
-        let memberProbe = self.testKit.makeTestProbe(expecting: SWIM.Message.self)
-        let ackProbe = self.testKit.makeTestProbe(expecting: SWIM.Message.self)
-        let payload: SWIM.GossipPayload = .membership([.init(peer: memberProbe.ref, status: .alive(incarnation: 0), protocolPeriod: 0)])
-        let pingReq: SWIM.Message = .remote(.pingRequest(target: memberProbe.ref, pingRequestOrigin: ackProbe.ref, payload: payload, sequenceNumber: 100))
-        try self.shared_serializationRoundtrip(pingReq)
-    }
+        guard let targetSwim = target._cluster?._swimShell else {
+            throw testKit.fail("SWIM shell should be non nil")
+        }
+        let targetID = ActorID(remote: target.settings.uniqueBindNode, path: targetSwim.id.path, incarnation: targetSwim.id.incarnation)
```
yim-lee, line 29: Have to do this for deserialized id to match

Sources/DistributedActors/Cluster/SWIM/SWIMActorShell.swift
```diff
@@ -527,6 +527,10 @@ internal distributed actor SWIMActorShell {
     func _getMembershipState() -> [SWIM.Member] {
         Array(self.swim.members)
     }
+    
+    nonisolated var description: String {
+        "\(Self.self)(\(self.id))"
+    }
```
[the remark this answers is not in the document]
ktoso: (the default description being that)

Tests/DistributedActorsTests/Cluster/SWIM/Protobuf/SWIM+SerializationTests.swift
ktoso: Aha I see since we need to force the remote...

There's a `._asRemote` I think that does this

Sources/DistributedActors/Cluster/SWIM/SWIMActorShell.swift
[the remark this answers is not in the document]
yim-lee: :nod

Tests/DistributedActorsTests/MetricsTestKit/MetricsTestKit.swift
```diff
@@ -379,6 +379,15 @@ public final class TestTimer: TestMetric, TimerHandler, Equatable, CustomStringC
...
 
+    public func reset() {
+        self.lock.withLock {
+            self._values = []
+        }
+        if self.verbose {
+            print("resetting \(self.label)")
+        }
+    }
+
```
ktoso, line 390: 👍

yim-lee:
Changed Samples to use Swift 5.7 instead of main.

Sources/DistributedActors/Cluster/SWIM/SWIMPeer+Actor.swift
```diff
@@ -18,16 +18,116 @@ import struct Dispatch.DispatchTime
 import enum Dispatch.DispatchTimeInterval
 import SWIM
 
+extension SWIMActorShell: SWIMAddressablePeer {
+    nonisolated var node: ClusterMembership.Node {
+        .init(protocol: self.id.uniqueNode.node.protocol, host: self.id.uniqueNode.host, port: self.id.uniqueNode.port, uid: self.id.uniqueNode.nid.value)
+    }
+}
+
+extension SWIMPeer {
```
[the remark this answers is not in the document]
yim-lee: Adding FIXME to code

Tests/DistributedActorsTests/Cluster/SWIM/SWIMActorShellClusteredTests.swift
```diff
@@ -13,442 +13,469 @@
...
-        first.cluster.join(node: second.cluster.uniqueNode.node)
-        try assertAssociated(first, withExactly: second.cluster.uniqueNode)
-        try assertAssociated(second, withExactly: first.cluster.uniqueNode)
+        // eventually it will ping/pingRequest and as all the pings (supposedly) time out it should mark as suspect
+        try await self.awaitStatus(.suspect(incarnation: 0, suspectedBy: [first.node]), for: targetPeer, on: first, within: .seconds(1))
+    }
 
-        let probeOnSecond = self.testKit(second).makeTestProbe(expecting: SWIM.Message.self)
-        let remoteMemberRef = first._resolveKnownRemote(probeOnSecond.ref, onRemoteSystem: second)
+    func test_swim_shouldNotMarkUnreachable_whenSuspectedByNotEnoughNodes_whenMinTimeoutReached() async throws {
```
yim-lee, line 299: This test needs "unresponsive test probe" to work. There are other tests in this file that would be better done with such test probe, but I manage to make them work with real systems.

yim-lee:
This PR is ready for review. Note that besides `SWIMActorShell`, there are major changes in the SWIM tests too.

Tests/DistributedActorsTests/Cluster/SWIM/SWIMActorShellClusteredTests.swift
ktoso: Right sorry I was working towards it but then uncovered the compiler blockers... Will it back shortly!

ktoso:
Wohoo! Going to have a look.

Sources/DistributedActors/Cluster/SWIM/SWIMActorShell.swift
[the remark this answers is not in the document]
ktoso: Yeah you're right -- it seems we're ending up making a lot of initializers async in other places as well, let's keep this  as is 👍

ktoso:
Validate against `main` swift-cluster-membership now

ktoso:
Okey let's keep progressing;

I'll focus on tests and hardening for a while now.

PR #980, opened 2022-07-01
Implement ActorSingleton for DistributedActor

yim-lee:
Resolves https://github.com/apple/swift-distributed-actors/issues/824

⚠️ WIP - haven't updated tests yet ⚠️


Sources/DistributedActors/ClusterSystem.swift
```diff
@@ -993,6 +993,10 @@ extension ClusterSystem {
         Err: Error,
         Res: Codable
     {
+        if let interceptor = actor.id.context.remoteCallInterceptor {
```
yim-lee, line 996: I wonder if we can move this inside `self.withCallID` so the interceptor doesn't need to worry about timeouts.

Sources/DistributedActors/Serialization/Serialization+Manifest.swift
```diff
@@ -131,7 +131,7 @@ extension Serialization {
     /// Tries to use getMangledTypeName if available.
     @inlinable
     @inline(__always)
-    internal static func getTypeHint(_ messageType: Any.Type) -> String {
+    public static func getTypeHint(_ messageType: Any.Type) -> String {
```
yim-lee, line 134: Will remove since we end up not serializing remote calls

Sources/DistributedActors/Serialization/Serialization.swift
```diff
@@ -782,6 +782,7 @@ public enum SerializationError: Error {
     case unableToMakeSerializer(hint: String)
     case unableToSerialize(hint: String)
     case unableToDeserialize(hint: String)
+    case unableToSummonType(hint: String)
```
yim-lee, line 785: Will remove since we end up not serializing remote calls

Sources/DistributedActors/ActorID.swift
```diff
@@ -100,7 +100,7 @@ extension ClusterSystem {
         public var tags: ActorTags
 
         /// Internal "actor context" which is used as storage for additional cluster actor features, such as watching.
-        internal var context: DistributedActorContext
+        public var context: DistributedActorContext
```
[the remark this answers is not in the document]
ktoso: I feel like we might either make this part of the main module... or make it _context somehow 🤔 

We really lost quite a bit of expressive power and "backdoors" with being fully in the language so we don't have as many sneaky ways around the normal swift rules as we did before...

Sources/ActorSingletonPlugin/ActorSingleton.swift
```diff
@@ -134,3 +52,57 @@ public enum AllocationStrategySettings {
...
+        Act.ID == ActorID,
+        Err: Error,
+        Res: Codable
+    {
+        // FIXME: better error handling
+        guard actor is Singleton else {
+            fatalError("Wrong interceptor")
+        }
+
+        // FIXME: can't capture inout param
```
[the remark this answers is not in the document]
ktoso: That's okey/expected I think -- we have to copy / capture by value the invocation... that sounds okey, since we'll store it for later use when we'd need to store and flush later.

Sources/ActorSingletonPlugin/ActorSingleton.swift
```diff
@@ -134,3 +52,57 @@ public enum AllocationStrategySettings {
...
+        let invocation = invocation
+        return try await self.proxy.whenLocal { __secretlyKnownToBeLocal in // TODO(distributed): this is annoying, we must track "known to be local" in typesystem instead
+            try await __secretlyKnownToBeLocal.forwardOrStashRemoteCall(target: target, invocation: invocation, throwing: throwing, returning: returning)
+        }! // FIXME: !-use
```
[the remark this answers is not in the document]
ktoso: > I don't know why the result is optional.

because at compile time we don't know that that actor is actually definitely local, so the signature of `whenLocal` is `-> T?`. 

So since the typesystem never knows and cannot remember "aha, this was `let x: local SomeActor`" we repeatedly and painfully keep reimplementing the whenLocal all the time when we do "framework-y stuff"... I expected this will be a pain and have been working on the `local` keyword but we don't have it yet 😢 

> I feel like I just switch, semi-automatically, to whenLocal whenever I see certain type of errors. Probably not good?

Kind of expected in "frameworky" code to be honest... That's why I kept mentioning in language work that we definitely want the `local` keyword, but it is not obvious why we need it so hard until one sees implementations like these where we're constantly forced to `whenLocal` all the time... 

In other words: I don't think it's your fault; it's expected and this is good signal and "proof" that we really need this local keyword as was expected tbh. This is good / useful, as basis / example why it matters so much.

I will review in depth tho to see if we're using it too much or not tho! I don't think it should be showing up much in "normal" user code, but in our frameworky code where we manage actors for users this is bound to show up a lot.

Sources/ActorSingletonPlugin/ActorSingletonPlugin.swift
```diff
@@ -113,18 +136,39 @@ public struct ActorSingletonControl {
...
 
-    /// Defines a singleton `behavior` and indicates that it can be hosted on this node.
-    public func host<Message>(_ type: Message.Type, name: String, props: _Props = _Props(), _ behavior: _Behavior<Message>) throws -> _ActorRef<Message> {
-        try self.singletonPlugin.ref(of: type, name: name, system: self.system, props: props, behavior)
+    /// Defines a singleton and indicates that it can be hosted on this node.
+    public func proxy<Act>(
+        _ type: Act.Type,
+        name: String,
+        props: _Props = _Props(),
+        _ factory: @escaping (ClusterSystem) async throws -> Act
```
ktoso, line 144: The "proxy" method means "always proxy but never host" so it would not have the factory function I think.

ktoso: We can revisit an discuss naming though

Sources/ActorSingletonPlugin/ActorSingletonProxy.swift
```diff
@@ -12,239 +12,387 @@
 //
 //===----------------------------------------------------------------------===//
```
ktoso, line 13: Funny filename actually -- the uppercase `I` there hah

Sources/DistributedActors/DistributedActorContext.swift
```diff
@@ -22,6 +22,8 @@ import Distributed
 public final class DistributedActorContext {
     let lifecycle: LifecycleWatchContainer?
 
+    public var remoteCallInterceptor: RemoteCallInterceptor?
```
ktoso, line 25: Would love to keep those internal hmmm

Sources/DistributedActors/Props.swift
```diff
@@ -74,7 +74,7 @@ public struct _Props: @unchecked Sendable {
 
     /// Allows for passing properties to creating a distributed actor.
     @TaskLocal
-    internal static var forSpawn: _Props = .init()
+    public static var forSpawn: _Props = .init()
```
ktoso, line 77: I don't think we can make it public, especially not without making `_underscored`.

We'll need something like it though... I'll think a bit how we want to expose it

Sources/DistributedActors/Props.swift
```diff
@@ -172,3 +172,17 @@ extension _Props {
...
+        props._knownActorName = name
+        return props
+    }
+    
+    public func _knownAs(name: String) -> Self {
+        var p = self
+        p._knownActorName = name
+        return p
+    }
+}
```
ktoso, line 188: okey as underscored 👍  the "known as name" stuff we neeed to get rid of but that's on me... okey to use it here.

Sources/ActorSingletonPlugin/ActorSingleton.swift
```diff
@@ -134,3 +52,57 @@ public enum AllocationStrategySettings {
...
+    ) async throws -> Res
+        where Act: DistributedActor,
+        Act.ID == ActorID,
+        Err: Error,
+        Res: Codable
+    {
+        // FIXME: better error handling
+        guard actor is Singleton else {
+            fatalError("Wrong interceptor")
+        }
```
ktoso, line 78: Hm... I don't mind the check, it feels like we should be able to guarantee by construction it is correct, but good to keep the check...

Sources/ActorSingletonPlugin/ActorSingletonProxy.swift
```diff
@@ -12,239 +12,387 @@
...
+    func forwardOrStashRemoteCall<Err, Res>(
+        target: RemoteCallTarget,
+        invocation: ActorSystem.InvocationEncoder,
+        throwing: Err.Type,
+        returning: Res.Type
+    ) async throws -> Res
+        where Err: Error,
+        Res: Codable
+    {
+        let singleton = await self.findSingleton()
```
ktoso, line 200: should this be throwing if we can't find a singleton within some timeout?

Sources/ActorSingletonPlugin/ActorSingletonProxy.swift
```diff
@@ -12,239 +12,387 @@
...
+        )
+    }
+
+    private func findSingleton() async -> Act {
+        await withCheckedContinuation { continuation in
+            // If singleton is available, forward remote call to it.
+            if let singleton = self.singleton {
+                continuation.resume(returning: singleton)
+                return
+            }
```
ktoso, line 236: this if can be moved out of the `withContinuation`

Sources/ActorSingletonPlugin/ActorSingletonProxy.swift
```diff
@@ -12,239 +12,387 @@
...
+            // Otherwise, we "stash" the remote call until singleton becomes available.
+            Task {
+                let callID = UUID()
+                self.log.debug("Stashing remote call [\(callID)]")
+                self.remoteCallContinuations.append((callID, continuation))
+                // FIXME: honor settings.bufferCapacity
+            }
```
ktoso, line 243: needs to also do some timeout task if we can't find the singleton on time?

Sources/ActorSingletonPlugin/ActorSingletonProxy.swift
```diff
@@ -12,239 +12,387 @@
...
+    static func _singleton(name: String, remote node: UniqueNode) -> ActorID {
+        ._make(remote: node, path: ._singleton(name: name), incarnation: .wellKnown)
     }
 }
 
 extension ActorPath {
-    static func _singletonProxy(name: String) -> ActorPath {
-        try! ActorPath._system.appending("singletonProxy-\(name)")
+    static func _singleton(name: String) -> ActorPath {
+        try! ActorPath._system.appending("singleton-\(name)")
```
ktoso, line 318: ok for now, we should move away from paths entirely though -- there's a new constructor on path which does not take a path and we should use tags to mark it as "this is a singleton" etc...

Dont have to do this in this PR tho, just FYI on direction

Sources/DistributedActors/ClusterSystem.swift
ktoso: Hmm I see, but it might have to do extra cleanup -- like this one does, when the stashed message times out, we'd like to drop it I think?

Sources/DistributedActors/Serialization/Serialization+Invocation.swift
```diff
@@ -21,14 +21,20 @@ import SwiftProtobuf
 
 public struct ClusterInvocationEncoder: DistributedTargetInvocationEncoder {
     public typealias SerializationRequirement = any Codable
-    var arguments: [Data] = []
+    public internal(set) var arguments: [Data] = []
```
ktoso, line 24: Hm this is because the cross module of the Singleton... I don't love exposing this as "really public" API... Need to think how to manage our cross module things. :-/

ktoso: specifically about this Data though we should be able to not serialize at all when the message is still stashed, OR may be delivered locally -- so we don';t do a serialization roundtrip... might need to have some two "modes" enum in the encoder.

I wish we were able to have just "different encoder per call" but we can't because compiler limitations: the ad-hoc protocol requirements strike again 😞 So we'll have to have one encoder type with "modes"

Sources/ActorSingletonPlugin/ActorSingletonManager.swift
```diff
@@ -12,77 +12,72 @@
...
-            case .stop:
-                // Hand over the singleton then stop myself as part of system shutdown
-                try self.handOver(context, to: nil)
-                return .stop
-            }
-        }
+    deinit {
+        // FIXME: should hand over
+        // TODO: perhaps we can figure out where `to` is next and hand over gracefully?
+//        try self.handOver(to: nil)
```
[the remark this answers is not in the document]
ktoso: Hmmm, okey interesting... will look deeper into this Q

Sources/ActorSingletonPlugin/ActorSingletonManager.swift
```diff
@@ -12,77 +12,72 @@
...
-    }
+    func takeOver(from: UniqueNode?) async throws -> Act {
+        self.log.debug("Take over singleton [\(self.settings.name)] from [\(optional: from)]", metadata: self.metadata())
 
-    private func handOver(_ context: _ActorContext<Directive>, to: UniqueNode?) throws {
-        // TODO: (optimization) tell `ActorSingletonManager` on `to` node that this node is handing off (https://github.com/apple/swift-distributed-actors/issues/329)
-        guard let singleton = self.singleton else {
-            return
+        // TODO: (optimization) tell `ActorSingletonManager` on `from` node that this node is taking over (https://github.com/apple/swift-distributed-actors/issues/329)
+        let singleton = try await _Props.$forSpawn.withValue(self.singletonProps._knownAs(name: self.settings.name)) {
```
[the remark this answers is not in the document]
ktoso: Yeah I know, this is a bit of a relic of old world as well as lacking new world API... 

Arguably naming "when spawning" isn't a thing at all in this new and "pure swift" world, and an actor shall have a name declared in itself -- I didn't get to this yet but I agree this entire API isn't user-facing at all by the way and must not be

Sources/ActorSingletonPlugin/ActorSingletonPlugin.swift
```diff
@@ -22,48 +23,70 @@ import DistributedActorsConcurrencyHelpers
 /// singleton running in the cluster.
 ///
 /// An actor singleton may run on any node in the cluster. Use `ActorSingletonSettings.allocationStrategy` to control
-/// its allocation. On candidate nodes where the singleton might run, use `ClusterSystem.singleton.ref(type:name:props:behavior)`
-/// to define actor behavior. Otherwise, call `ClusterSystem.singleton.ref(type:name:)` to obtain a ref. The returned
-/// `_ActorRef` is in reality a proxy which handle situations where the singleton is shifted to different nodes.
+/// its allocation. On candidate nodes where the singleton might run, use `ClusterSystem.singleton.proxy(type:name:system:props:factory)`
```
ktoso, line 26: So "proxy" is the wrong name here; since this is the API where "we might end up hosting it" and I don't want to always talk about proxies in APIs...

I think we should strive for an API like:

```swift
let singleton: ClusterSingleton(..., actorSystem: system)
await singleton.get(SomeActor.self)
```

perhaps? 

We can think if we want the `SingletonBoss<DA>` or just `SingletonBoss` hm, first might be easier to manage and implement? less type casting <3

Sources/ActorSingletonPlugin/ActorSingletonPlugin.swift
```diff
@@ -22,48 +23,70 @@ import DistributedActorsConcurrencyHelpers
...
+    {
+        if let existingID = self.singletons[settings.name]?.proxied {
+            return try Act.resolve(id: existingID, using: system)
+        }
 
-            guard let proxy = singleton.proxy else {
-                fatalError("Singleton[\(settings.name)] not yet initialized")
-            }
+        // Spawn the proxy for the singleton (one per singleton per node)
+        let proxy = try await _Props.$forSpawn.withValue(_Props._wellKnownActor(name: "singletonProxy-\(settings.name)")) {
```
ktoso, line 52: TODO: rework this entire spawning / props pattern -- I'll work on this as soon as In can this week

Sources/ActorSingletonPlugin/ActorSingletonPlugin.swift
```diff
@@ -22,48 +23,70 @@ import DistributedActorsConcurrencyHelpers
...
+                system: system,
+                singletonProps: props,
+                factory
+            )
+        }
 
-            return proxy
+        // Assign the singleton actor a special id and set up the remote call interceptor
+        var id = _Props.$forSpawn.withValue(_Props._wellKnownActor(name: "singleton-\(settings.name)")) {
+            system.assignID(Act.self)
```
ktoso, line 63: TODO: note to self that this entire well known is not so great with DA and need to get away from this props via task local pattern here. 

I'll think about what we want to do here, but in general I think we should probably expose a new `assignID` to the system which takes various options how it assigns it... and this way we can use it from the singleton module, without these task-local trickery in here (in this specific instance we don't have to do it this way)

I'll follow up here with a patch to show what we could do so we can check it :)

Sources/ActorSingletonPlugin/ActorSingletonPlugin.swift
```diff
@@ -73,16 +96,16 @@ extension ActorSingletonPlugin {
...
         Self.pluginKey
     }
 
     public func start(_ system: ClusterSystem) async throws {}
 
-    public func stop(_ system: ClusterSystem) {
-        self.singletonsLock.withLock {
-            for (_, singleton) in self.singletons {
-                singleton.stop(system)
+    public nonisolated func stop(_ system: ClusterSystem) {
```
ktoso, line 105: Arguably should probably become async instead?

ktoso:
Okey I think I have some work to be done here to make it nicer to implement! I'll follow up as soon as I'm out of compiler land <3

Sources/DistributedActors/ActorID.swift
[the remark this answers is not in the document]
yim-lee: Yeah, there are a few things in `_Props` as well and luckily (?) `remoteCall` and `remoteCallVoid` are already public. I would rather move singleton into the main module. wdyt?

Sources/ActorSingletonPlugin/ActorSingletonPlugin.swift
yim-lee: > "always proxy but never host"

Right. I can't recall why we have this mode though in the original impl. Does it make sense to have it?

Sources/DistributedActors/DistributedActorContext.swift
yim-lee: Think so, just need to update `init`.

Sources/ActorSingletonPlugin/ActorSingletonProxy.swift
yim-lee: Hmm??  o_O

Sources/ActorSingletonPlugin/ActorSingleton.swift
yim-lee: Yeah, I find it weird having this check, and something is seriously wrong if we ever got here and actor is of a different type, but then it also doesn't feel right without the check. Dunno...I might remove this in the end.

Sources/ActorSingletonPlugin/ActorSingletonPlugin.swift
yim-lee: The DA impl doesn't have `ActorSingleton` anymore so yes, we can rename this to `ClusterSingleton` I think. 

I also thought about removing manager/boss because I don't think we really need it...

Sources/ActorSingletonPlugin/ActorSingletonPlugin.swift
yim-lee: Yes, though TBH not sure how to handle async in `ClusterSystem.shutdown`.

Sources/ActorSingletonPlugin/ActorSingleton.swift
ktoso: Keeping is fine, just in case™

ktoso:
I'll check this out and see how to improve use of actor context and metadata now 👍

Sources/ActorSingletonPlugin/ActorSingletonPlugin.swift
yim-lee: I renamed things in https://github.com/apple/swift-distributed-actors/pull/980/commits/bfc125a1ccb8f47a0ea1cd8d9a175f63c0cc8cb4. Hopefully it's cleaner now.

Sources/ActorSingletonPlugin/ActorSingletonProxy.swift
yim-lee: I reworked this bit in https://github.com/apple/swift-distributed-actors/pull/980/commits/1de268f71ab7fca8c36fec52a3a557d5c18155d0

I guess only the actual singleton DA needs to have "known name". `ClusterSingleton` (previously `ActorSingletonProxy`) and `ClusterSingletonBoss` (`ActorSingletonManager`) don't need to be looked up, so they can be named whatever. The singleton DA though, needs to be well-known so we can construct `ActorID` for it. For now the code still uses `knownActorName` and `path`, but maybe we want to do things differently now that we have `ActorMetadata`.

ktoso: I'll dig into this -- the "known" won't work at all once we remove paths since the "known" is just "id == 0" and we cannot have more of those -- previously we could since in the akka and path model the identity is "path + id" but now it is only id...

SO! This is why the metadata was so important, now we can have the metadata participate in identity and this way we can still do "singleton is well known" with a special tag about it. I'll figure this out today into this PR 👍

Tests/DistributedActorsTests/ClusterSystemTests.swift
```diff
@@ -378,29 +380,80 @@ final class ClusterSystemTests: ClusterSystemXCTestCase {
...
+        let remote = await setUpNode("remote") { settings in
+            settings.enabled = true
+        }
+        local.cluster.join(node: remote.cluster.uniqueNode)
+
+        let greeter = Greeter(actorSystem: local, greeting: "hello")
+        let localGreeter = try Greeter.resolve(id: greeter.id, using: remote)
+
+        let otherGreeter = Greeter(actorSystem: local, greeting: "HI!!!")
+        localGreeter.id.context.remoteCallInterceptor = GreeterRemoteCallInterceptor(system: remote, greeter: otherGreeter)
```
ktoso, line 397: We can't do this -- context must not be mutated from the outside the actor.

ktoso: (For context here's some new actor system APIs to enable this code without assigning to context from outside https://github.com/yim-lee/swift-distributed-actors/commit/aa73aa1f6a2cf643c293306aa28f987d725b619d )

Sources/ActorSingletonPlugin/ActorSingletonPlugin.swift
ktoso: Followed up

Sources/ActorSingletonPlugin/ActorSingletonPlugin.swift
ktoso: Yup nicer

Sources/DistributedActors/ClusterSystem.swift
```diff
@@ -986,7 +986,7 @@ extension ClusterSystem {
 // MARK: Intercepting calls
 
 extension ClusterSystem {
-    public func interceptCalls<Act, Interceptor>(
+    internal func interceptCalls<Act, Interceptor>(
```
ktoso, line 989: great, thanks -- I was a bit nervous if it is the right exact API we want to support forever.

I'm pretty sure we'll expose this eventually, but let's buy us some time :)

Sources/DistributedActors/ClusterSystem.swift
```diff
@@ -1051,7 +1051,6 @@ extension ClusterSystem {
         }
 
         if let error = reply.thrownError {
-            print("error: \(error)")
```
ktoso: uf, sorry!

Sources/DistributedActors/ActorID.swift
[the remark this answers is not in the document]
ktoso: (we moved :))

Sources/DistributedActors/Plugins/ClusterSingleton/ClusterSingleton.swift
```diff
@@ -13,33 +13,31 @@
...
 import struct Foundation.Data
 import struct Foundation.UUID
 import Logging
 
 // ==== ----------------------------------------------------------------------------------------------------------------
-// MARK: Cluster singleton
+// MARK: Cluster singleton boss
 
-internal protocol ClusterSingletonProtocol {
+internal protocol _ClusterSingletonBoss {
```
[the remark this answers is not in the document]
yim-lee: Yeah I will rename it to `ClusterSingletonBossProtocol`. There was a name conflict. :)

Sources/DistributedActors/ClusterSystem.swift
yim-lee: np at all!

Sources/DistributedActors/Plugins/ClusterSingleton/ClusterSingletonPlugin.swift
```diff
@@ -14,23 +14,16 @@
...
 import Distributed
 
-extension ActorMetadataKeys {
-    public var clusterSingletonID: Key<String> { "$singleton-id" }
-}
-
-public protocol ClusterSingletonProtocol: DistributedActor where ActorSystem == ClusterSystem {
-    /// Must be implemented using a `@ActorID.Metadata(\.clusterSingletonID)` annotated property.
-    var singletonName: String { get }
-}
```
ktoso: yup right, we wont be using this after all

Sources/DistributedActors/Plugins/ClusterSingleton/ClusterSingleton.swift
```diff
@@ -271,21 +282,85 @@ internal distributed actor ClusterSingletonBoss<Act: DistributedActor>: _Cluster
...
+                case .none:
+                    self.status = .pending
+                    if self.timeoutTask == nil {
+                        Task {
+                            await self.startTimeoutTask()
+                        }
+                    }
+                }
+            }
+        }
```
ktoso, line 305: okey good -- why as separate actor though, the boss could do this inside it already or do I miss something?

We'll get an extra actor hop when interacting with this one and issuing calls, preferably we wouldn't hm

Sources/DistributedActors/Plugins/ClusterSingleton/ClusterSingleton.swift
```diff
@@ -109,7 +110,8 @@ internal distributed actor ClusterSingletonBoss<Act: DistributedActor>: _Cluster
 
     private func updateTargetNode(node: UniqueNode?) async throws {
         guard self.targetNode != node else {
-            self.log.debug("Skip updating target node. New node is already the same as current targetNode.", metadata: self.metadata())
+            let metadata = await self.metadata()
```
ktoso, line 113: oh wow this is bad, definitely not async logger metadata access. (Why it's bad: this will always render all the metadata, AND causes suspensions even if we're not even logging -- the logging must be done inside the `debug(...)` call because it is lazily evaluated (metadata is an auto closure)

It's because the move of the singleton in to the other actor -- `if let singleton = await self.singleton {`
but why did we do this in a new actor to begin with...? This actor should be able to do everything necessary?

yim-lee: amended https://github.com/apple/swift-distributed-actors/pull/980/commits/3b882d62eb53c543108e291f2ffdab95105224de

Sources/DistributedActors/Plugins/ClusterSingleton/ClusterSingleton.swift
yim-lee: amended https://github.com/apple/swift-distributed-actors/pull/980/commits/3b882d62eb53c543108e291f2ffdab95105224de

yim-lee:
Make this PR mergeable to prepare for https://github.com/apple/swift-distributed-actors/pull/1001

Sources/DistributedActors/Cluster/Downing/DowningStrategy.swift
```diff
@@ -118,11 +118,12 @@ internal distributed actor DowningStrategyShell {
             self.memberTimerTasks[member] = Task {
                 defer { self.memberTimerTasks.removeValue(forKey: member) }
 
+                try await Task.sleep(until: .now + delay, clock: .continuous)
+
```
ktoso, line 122: ah nice thx

Sources/DistributedActors/ClusterSystem.swift
```diff
@@ -493,7 +493,12 @@ public class ClusterSystem: DistributedActorSystem, @unchecked Sendable {
         /// Down this node as part of shutting down; it may have enough time to notify other nodes on an best effort basis.
         self.cluster.down(node: self.settings.node)
 
-        self.settings.plugins.stopAll(self)
+        let pluginsSemaphore = DispatchSemaphore(value: 1)
+        Task {
+            await self.settings.plugins.stopAll(self)
+            pluginsSemaphore.signal()
+        }
+        pluginsSemaphore.wait()
```
ktoso, line 501: another case of "we need `send self settings.plugins.stopAll(self)`" strikes again huh...

ktoso: Ah arguably we could make the shutdown `async` now 🤔

Sources/DistributedActors/Plugins/ClusterSingleton/ClusterSingleton.swift
```diff
@@ -0,0 +1,448 @@
...
+///
+/// `ClusterSingletonBoss` has a buffer to hold remote calls temporarily in case the singleton is not available.
+/// The buffer capacity is configurable in `ClusterSingletonSettings`. Note that if the buffer becomes
+/// full, the *oldest* message would be disposed to allow insertion of the latest message.
+///
+/// `ClusterSingletonBoss` subscribes to cluster events and feeds them into `AllocationStrategy` to
+/// determine the node that the singleton runs on. If the singleton falls on *this* node, `ClusterSingletonBoss`
+/// will spawn the actual singleton actor. Otherwise, `ClusterSingletonBoss` will hand over the singleton
+/// whenever the node changes.
+internal distributed actor ClusterSingletonBoss<Act: ClusterSingletonProtocol>: ClusterSingletonBossProtocol where Act.ActorSystem == ClusterSystem {
```
ktoso, line 40: (love the boss naming, makes me smile each time lol) Enough manager object's already! 😆

Sources/DistributedActors/Plugins/ClusterSingleton/ClusterSingleton.swift
```diff
@@ -0,0 +1,448 @@
...
+        if system.settings.enabled {
+            self.clusterEventsSubscribeTask = Task {
+                // Subscribe to ``Cluster/Event`` in order to update `targetNode`
+                for await event in system.cluster.events {
+                    try await self.receiveClusterEvent(event)
+                }
+            }
+        } else {
+            // Run singleton on this node if clustering is not enabled
+            self.log.debug("Clustering not enabled. Taking over singleton.")
```
ktoso, line 90: ah, nicely handled as well

Sources/DistributedActors/Plugins/ClusterSingleton/ClusterSingleton.swift
```diff
@@ -0,0 +1,448 @@
...
+        // TODO: (optimization) tell `to` node that this node is handing off (https://github.com/apple/swift-distributed-actors/issues/329)
+        self.updateSingleton(nil)
+    }
+
+    private func updateSingleton(node: UniqueNode?) throws {
+        switch node {
+        case .some(let node) where node == self.actorSystem.cluster.uniqueNode:
+            ()
+        case .some(let node):
+            let singleton = try Act.resolve(id: .singleton(Act.self, settings: self.settings, remote: node), using: self.actorSystem)
```
ktoso, line 159: Just FYI, That's one of the places I have new impl based on well-known metadata for :)

Sources/DistributedActors/Plugins/ClusterSingleton/ClusterSingleton.swift
```diff
@@ -0,0 +1,448 @@
...
+                        self.log.warning("Unable to stash remote call, error: \(error)", metadata: self.metadata())
+                        continuation.resume(throwing: ClusterSingletonError.stashFailure)
+                    }
+                }
+            }
+        }
+    }
+
+    nonisolated func stop() async {
+        Task {
```
ktoso, line 261: Could do without this outer task I think, will try locally

Sources/DistributedActors/Plugins/ClusterSingleton/ClusterSingleton.swift
ktoso: Great thanks!

Sources/DistributedActors/Plugins/ClusterSingleton/ClusterSingleton.swift
```diff
@@ -0,0 +1,448 @@
...
+                return nil
+            }
+            return self.buffer.removeFirst()
+        }
+    }
+
+    enum BufferError: Error {
+        case full
+    }
+}
```
ktoso, line 331: 👍 Basically our new "async" stash replacement, looking good

Tests/DistributedActorsTests/Plugins/ClusterSingleton/ClusterSingletonPluginTests.swift
```diff
@@ -0,0 +1,41 @@
...
+
+        // singleton.host
+        let ref = try await test.singleton.host(of: TheSingleton.self, name: TheSingleton.name) { actorSystem in
+            TheSingleton(greeting: "Hello", actorSystem: actorSystem)
+        }
+        let reply = try await ref.greet(name: "Charlie")
+        reply.shouldStartWith(prefix: "Hello Charlie!")
+
+        // singleton.ref (proxy-only)
+        let proxyRef = try await test.singleton.proxy(of: TheSingleton.self, name: TheSingleton.name)
```
ktoso, line 37: Impl wise I actually realized that we don't handle this well, we might just do the host version for now.

yim-lee: 👌

PR #981, opened 2022-07-05
fix wrong cross module usage; seems this was not caught by CI / SwiftPM

ktoso:
Suprisingly this seems to have been building fine on CI, but triggeres errors in Xcode -- where I suppose the module checks are actually enforced...? I'll dig into the actual problems but we can fix this right away anyway.


ktoso:
Related SwiftPM issue / PR which would have prevented such issues: https://github.com/apple/swift-package-manager/pull/3562

Issue #983, opened 2022-07-05
Attaching region metadata to a distributed actor


PopFlamingo:
Some use cases might require to be aware of the region a specific actors lives in, for instance in distributed matchmaking where region preferences matter for delays or even networking costs for the host.

I was wondering what is the best way to attach such information to actors of the cluster system?

Thank you

ktoso:
I like the example case of "region", thanks for bringing it up :)

We're making ways to attach metadata to an `ActorID` public now, so I think that'd address this... Will link the related PR here soon

PopFlamingo:
@ktoso Thank you, awesome!

ktoso:
I'll add some more docs and integrate with receptionist soon as well 👍

Issue #479, opened 2020-03-02
`ActorID` rather than "Unique ActorPath" for resolve and identity


ktoso:
This relates to the big "do not use paths for core of identity, but just ActorIDs".

Old ticket was:

----

This is in remoting since it impacts how we do `resolve(name) -> Ref`

----

Our current paths design is inherited from Akka:

- Actor HAS-A (Unique)ActorPath
- identity is carried by the (Unique)ActorPath

This matters for remoting actually, since:

**Sending ActorRefs:**

- we need to convert a ref to its "distributable format;
  - in Akka this is the string representation of the path; `akka://system@10.0.01:5555/user/example/something#UID`
  - downsides:
    - path rendering is expensive-ish
    - path as String means envelope has to be able to carry variable-length here
- the ref while local does not always carry the host part, and needs to be populated when rendering
  - this needs to work with codable

**Receiving ActorRefs:**
- Upon receiving an `akka://system@10.0.01:5555/user/example/something#UID` we have to check:
  - is the destination system "me"? if yes then continue:
    - in reality, the system also carries its own UUID since we have to know if it is the same instance of a system on this address, or a new one (!)
  - we need to transform the `/user/example/something#UID` path into the actual `ActorRef` which inserts messages to the right mailbox
    - this is done by some "lookup", we call this "resolve the actor ref" kind of like dns resolvers -- `name -> actual thing` after all
    - if the `Path` is the source of truth the implementation to resolve is: **traverse the actor tree from top through all children using the path, and once leaf reached verify is is the right UUID** (since diff actor could have started on the same path, but it's different and shall not receive this message
       - if the actor is not resolved, the message is "dead-lettered", since the actor seems to have died before the message reached it

The not so nice part is the traversing, since all access to `.children` needs to be lock protected (well, or atomic dances to immutable structures). and the traversal grows in time as deeper the children become (each level being one dictionary lookup by name).

Akka is Path as identifier focused because historically it allowed `actorSelection(String): ActorRef` which is NOT something that we want (or _can_) offer since the type of the actor ref is not known! This is solved by other mechanisms.

## Alternative design

Key changes: 
- **Paths are only for humans.**
- `AID` -- like a PID but for actors.
- registry allowing to look up AID to `ReceivesSystemMessages`.
  - TODO: actor deaths need to remove from the registry

[...]

Issue #979, opened 2022-06-23
Better docs around "find actor by 'user id'"


ktoso:
For the simplest case, using existing API, we're able to do this with receptionist and checking in with "user id" reception keys.
It is somewhat sub optimal, but that is one way to do it manually today.

Future directions include using singletons to do that, or sharding/virtual actors which would allow that.

PR #988, opened 2022-07-06
Implement setting actor metadata with property wrapper fields

ktoso:
Resolves https://github.com/apple/swift-distributed-actors/issues/479
Resolves https://github.com/apple/swift-distributed-actors/issues/983
Provides a good usability basis for https://github.com/apple/swift-distributed-actors/issues/979

This allows for the following:

```swift
    distributed actor Example: CustomStringConvertible {
        
        @ActorID.Metadata(\.exampleUserID)
        var userID: String

        init(userID: String, actorSystem: ActorSystem) async {
            self.actorSystem = actorSystem
            self.userID = userID
        }
    }
```

```swift
        let userID = "user-1234"
        let example = await Example(userID: userID, actorSystem: system)
        
        example.id.metadata.exampleUserID.shouldEqual(userID)
```

and replaces our `_Props.$forSpawn.withValue(...) { ... }` dance.

I've not replaced all the uses of the forSpawn yet, but we should be able to for all DA declarations we have 👍

### DIFF
--- a/Sources/DistributedActors/ActorID.swift
+++ b/Sources/DistributedActors/ActorID.swift
@@ -509,10 +509,6 @@ extension ActorPath {
     public static let _user: ActorPath = try! ActorPath(root: "user")
     public static let _system: ActorPath = try! ActorPath(root: "system")
 
-    // Since in a pure DA world, we don't rely on paths, this path may sometimes be returned during migration period
-    // to indicate that an ActorID does not have a path defined.
-    public static let _undefined: ActorPath = try! ActorPath(root: "undefined")
-
     internal func makeLocalID(on node: UniqueNode, incarnation: ActorIncarnation) -> ActorID {
         .init(local: node, path: self, incarnation: incarnation)
     }
@@ -905,9 +901,9 @@ extension UniqueNodeID {
 
 extension ActorID: Codable {
     public func encode(to encoder: Encoder) throws {
-        let tagSettings = encoder.actorSerializationContext?.system.settings.tags
-        let encodeCustomTags: (ActorID, inout KeyedEncodingContainer<ActorCoding.TagKeys>) throws -> Void =
-            tagSettings?.encodeCustomTags ?? ({ _, _ in () })
+        let metadataSettings = encoder.actorSerializationContext?.system.settings.actorMetadata
+        let encodeCustomMetadata: (ActorID, inout KeyedEncodingContainer<ActorCoding.MetadataKeys>) throws -> Void =
+            metadataSettings?.encodeCustomMetadata ?? ({ _, _ in () })
 
         var container = encoder.container(keyedBy: ActorCoding.CodingKeys.self)
         try container.encode(self.uniqueNode, forKey: ActorCoding.CodingKeys.node)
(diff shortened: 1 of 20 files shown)


Sources/DistributedActors/ActorTagSettings.swift
```diff
@@ -43,8 +43,8 @@ public struct ActorTagSettings {
     /// What type of tags, known and defined by the cluster system itself, should be automatically propagated.
     /// Other types of tags, such as user-defined tags, must be propagated by declaring apropriate functions for `encodeCustomTags` and `decodeCustomTags`.
     internal var propagateTags: Set<AnyActorTagKey> = [
-        .init(ActorTags.path),
-        .init(ActorTags.type),
+        .init(ActorMetadata.path),
+        .init(ActorMetadata.type),
     ]
```
yim-lee, line 48: Maybe a cleanup task for later is to rename tag related types and properties to metadata?

Sources/DistributedActors/DistributedActorContext.swift
```diff
@@ -12,6 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Dispatch
```
yim-lee, line 15: used?

Sources/DistributedActors/DistributedActorContext.swift
```diff
@@ -21,18 +22,22 @@ import Distributed
...
 
-    init(lifecycle: LifecycleWatchContainer?) {
+    init(lifecycle: LifecycleWatchContainer?,
+         metadata: ActorMetadata? = nil)
+    {
         self.lifecycle = lifecycle
-        traceLog_DeathWatch("Create context; Lifecycle: \(optional: lifecycle)")
+        self.metadata = metadata ?? ActorMetadata()
+
+        traceLog_DeathWatch("Create context; Lifecycle: \(lifecycle)")
```
yim-lee, line 33: Need `optional` or `String(describing:)` around `lifecycle`?

Sources/DistributedActors/ActorMetadata.swift
```diff
@@ -12,44 +12,77 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Dispatch
```
yim-lee, line 15: used?

Sources/DistributedActors/ActorID.swift
```diff
@@ -197,17 +231,27 @@ extension ClusterSystem {
...
-            // TODO: avoid mangling names on every spawn?
-            if let mangledName = _mangledTypeName(type) {
-                self.tags[ActorTags.type] = .init(mangledName: mangledName)
+            if let mangledName = _mangledTypeName(type) { // TODO: avoid mangling names on every spawn?
+                self.context.metadata[ActorMetadata.type] = .init(mangledName: mangledName)
             }
             traceLog_DeathWatch("Made ID: \(self)")
         }
 
         internal var withoutContext: Self {
```
yim-lee, line 240: Should this be renamed to `withoutLifecycle`? Or make `metadata` `nil` as well?

Sources/DistributedActors/ActorTagSettings.swift
ktoso: Good one, will do right away

Sources/DistributedActors/DistributedActorContext.swift
ktoso: right thanks :)

Sources/DistributedActors/ActorID.swift
ktoso: Good one, thanks -- it kind of is "without stuff ... but keep the tags" hm, ill rename for now

Sources/DistributedActors/ActorTagSettings.swift
ktoso: done

ktoso:
Sigh was happy all day and now a bit flaky...

Issue #943, opened 2022-06-05
Dead letter when a message arrives from remote call and target is not alive anymore


ktoso:
In DA world this only happens for distributed remote calls, it used to be possible everywhere and we implemented a "personal dead letters" for it.

Instead, we now must throw during `.resolve` and catch during applying a remote call -- and cause a dead letter there.

Include logging tests so we make sure it actually did dead letter.

Relevant pieces of code:

```
            guard let managed = self._managedDistributedActors.get(identifiedBy: address) else {
                log.trace("Resolved as remote reference", metadata: [
                    "actor/identity": "\(address)",
                ])
                // TODO(distributed): throw here, this should be a dead letter
                return nil
            }
```

in `resolve` and `receiveInvocation` is where we should catch and cause the dead letter.

PR #989, opened 2022-07-07
Dead letter when remote call target is dead

yim-lee:
Resolves https://github.com/apple/swift-distributed-actors/issues/943


Sources/DistributedActors/ClusterSystem.swift
```diff
@@ -876,13 +876,12 @@ extension ClusterSystem {
             return nil
         }
 
-        return self.namingLock.withLock {
+        return try self.namingLock.withLock {
             guard let managed = self._managedDistributedActors.get(identifiedBy: id) else {
```
yim-lee, line 880: IIUC, only local resolve reaches here?

Sources/DistributedActors/ClusterSystem.swift
```diff
@@ -1148,6 +1142,12 @@ extension ClusterSystem {
             )
 
             do {
+                guard let actor = self.resolve(id: recipient) else {
+                    self.log.error("Unable to resolve recipient \(recipient). Message will be dropped: \(invocation)")
+                    self.deadLetters.tell(DeadLetter(invocation, recipient: recipient))
```
yim-lee, line 1147: Should the dead letter happen at sender, after getting back `DeadLetterError`? Or both sender and recipient? 🤔

ktoso: On the recipient I believe, since the recipient will has no function that will throw, so this dead letter informs that system that "someone is calling things on you, but they can't get through to the recipient actor".

And the caller has the function that they called that resulted in a dead letter, that function should _throw_ as it received s the error reply from the dead-letter-ed remoteCall. This way the caller knows it happened, we don't need to log it there.

Sources/DistributedActors/ClusterSystem.swift
```diff
@@ -1148,6 +1142,12 @@ extension ClusterSystem {
             )
 
             do {
+                guard let actor = self.resolve(id: recipient) else {
+                    self.log.error("Unable to resolve recipient \(recipient). Message will be dropped: \(invocation)")
```
ktoso, line 1146: Only sent it to dead letters, no need to log here -- dead letters will do the logging 👍 
```suggestion

```

Sources/DistributedActors/DeadLetters.swift
```diff
@@ -12,6 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Distributed
+import Foundation
```
ktoso, line 16: foundation not used here, Codable is stdlib :) 
```suggestion

```

Sources/DistributedActors/DeadLetters.swift
```diff
@@ -255,3 +257,10 @@ extension ActorPath {
     static let _dead: ActorPath = try! ActorPath(root: "dead")
     static let _deadLetters: ActorPath = try! ActorPath._dead.appending("letters")
 }
+
+// ==== ----------------------------------------------------------------------------------------------------------------
+// MARK: Errors
+
+public struct DeadLetterError: DistributedActorSystemError, Codable {
+    public let recipient: ClusterSystem.ActorID
+}
```
ktoso, line 266: 👍

Tests/DistributedActorsTests/DeadLetterTests.swift
```diff
@@ -80,4 +81,82 @@ final class DeadLetterTests: ClusterSystemXCTestCase {
...
+
+        var greeter: Greeter? = Greeter(actorSystem: local)
+        let localGreeter = try Greeter.resolve(id: greeter!.id, using: remote)
+
+        let p = self.testKit.makeTestProbe(expecting: String.self)
+        let watcher = GreeterWatcher(probe: p, actorSystem: local)
+        try await watcher.watch(greeter!)
+
+        greeter = nil
+        try p.expectMessage(prefix: "Received terminated: /user/Greeter")
```
ktoso, line 102: Oh we're missing API to `p.watch` a DA from a TestProbe it seems? If you want that'd be a great thing to implement...

We'd want to be able to `p.watch(watcher); p.expectTerminated()`

ktoso (approved):
Tests and impl look good! Just a minor nitpick about not logging in one place.

LGTM, please merge at will 👍

Tests/DistributedActorsTests/DeadLetterTests.swift
yim-lee: Sure. Created https://github.com/apple/swift-distributed-actors/issues/990

Issue #990, opened 2022-07-07
Add ability to watch distributed actor with TestProbe


yim-lee:
https://github.com/apple/swift-distributed-actors/pull/989#discussion_r915885909

Goal:

```
p.watch(watcher)
p.expectTerminated()
```

Issue #945, opened 2022-06-06
FAILED: ClusterLeaderActionsClusteredTests.test_singleLeader


ktoso:
Maybe just flaky, but happened on updating NIO so maybe timings changed and the test was too sensitive to them?

https://ci.swiftserver.group/job/swift-distributed-actors-nightly-prb/357/consoleFull

```
18:22:46 Test Case 'ClusterLeaderActionsClusteredTests.test_singleLeader' started at 2022-06-05 09:22:46.980
18:22:52 [pprint][9:22:46.9840] [/code/Sources/DistributedActors/Cluster/ClusterShell.swift:271][thread:140641564575488]: Started ClusterShell
18:22:52 <EXPR>:0: error: ClusterLeaderActionsClusteredTests.test_singleLeader : threw error "
18:22:52         switch try p.expectMessage() {
18:22:52                     ^~~~~~~~~~~~~~
18:22:52 error: Did not receive message of type [Event] within [5s], error: noMessagesInQueue"
18:22:52 ------------------------------------- ClusterSystem(first, sact://first@127.0.0.1:7111) ------------------------------------------------
18:22:52 Captured log [first][2022-06-05 9:22:46.9850] [ClusterSystem.swift:894] [warning] Assign identity
18:22:52 // metadata:
18:22:52 // "actor/id": /system/receptionist
18:22:52 // "actor/id/uniqueNode": sact://first@127.0.0.1:7111
18:22:52 // "actor/type": OpLogDistributedReceptionist
18:22:52 Captured log [first][2022-06-05 9:22:46.9850] [ClusterShell.swift:409][/system/cluster] [info] Binding to: [sact://first@127.0.0.1:7111]
18:22:52 Captured log [first][2022-06-05 9:22:46.9850] [ClusterSystem.swift:907] [trace] Actor ready
18:22:52 // metadata:
18:22:52 // "actor/id": /system/receptionist
18:22:52 // "actor/type": OpLogDistributedReceptionist
18:22:52 Captured log [first][2022-06-05 9:22:46.9860] [Leadership.swift:114][/system/cluster/leadership] [trace] Configured with LowestReachableMember(minimumNumberOfMembersToDecide: 1, loseLeadershipIfBelowMinNrOfMembers: false)
18:22:52 Captured log [first][2022-06-05 9:22:46.9860] [ClusterSystem.swift:894] [warning] Assign identity
18:22:52 // metadata:
18:22:52 // "actor/id": /system/downingStrategy
[...]

PR #991, opened 2022-07-08
Add ability to watch distributed actor with TestProbe

yim-lee:
Resolves https://github.com/apple/swift-distributed-actors/issues/990

### DIFF
--- a/Sources/DistributedActorsTestKit/TestProbes.swift
+++ b/Sources/DistributedActorsTestKit/TestProbes.swift
@@ -805,7 +805,7 @@ extension ActorTestProbe {
         return watchee
     }
 
-    /// Checks that the actor has been terminated.
+    /// Suspends waiting for the actor to become terminated.
     ///
     /// - Warning: Remember to first `watch` the actor you are expecting termination for,
     ///            otherwise the termination signal will never be received.
--- a/Sources/DistributedActorsTestKit/TestProbes.swift
+++ b/Sources/DistributedActorsTestKit/TestProbes.swift
@@ -810,9 +810,8 @@ extension ActorTestProbe {
     ///            otherwise the termination signal will never be received.
     ///
     /// - SeeAlso: `DeathWatch`
-    // TODO: expectTermination(of: ...) maybe nicer wording?
-    public func expectTerminated(
-        _ actor: ActorID,
+    public func expectTermination(
+        of actor: ActorID,
         within timeout: Duration? = nil,
         file: StaticString = #filePath, line: UInt = #line, column: UInt = #column
     ) async throws {
--- a/Sources/DistributedActorsTestKit/TestProbes.swift
+++ b/Sources/DistributedActorsTestKit/TestProbes.swift
@@ -12,6 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Distributed
 @testable import DistributedActors
 import DistributedActorsConcurrencyHelpers
 import Foundation
@@ -47,6 +48,8 @@ public final class ActorTestProbe<Message: Codable>: @unchecked Sendable {
     internal let internalRef: _ActorRef<ProbeCommands>
     internal let exposedRef: _ActorRef<Message>
 
+    private lazy var _internal = _TestProbeInternal(actorSystem: self.system)
+
     /// The reference to the underlying "mock" actor.
     /// Sending messages to this reference allows the probe to inspect them using the `expect...` family of functions.
     public var ref: _ActorRef<Message> {
@@ -57,6 +60,8 @@ public final class ActorTestProbe<Message: Codable>: @unchecked Sendable {
     private var expectationTimeout: Duration {
         self.settings.expectationTimeout
     }
+
+    private let system: ClusterSystem
 
     /// Blocking linked queue, available to run assertions on
     private let messagesQueue = _LinkedBlockingQueue<Message>()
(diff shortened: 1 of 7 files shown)


Sources/DistributedActorsTestKit/TestProbes.swift
```diff
@@ -58,6 +61,8 @@ public final class ActorTestProbe<Message: Codable>: @unchecked Sendable {
         self.settings.expectationTimeout
     }
 
+    private let system: ClusterSystem
```
ktoso, line 64: do we need to store the system, or just initialize the internal test probe with it is enough? Would be good to not store the system if we can hm

Sources/DistributedActorsTestKit/TestProbes.swift
```diff
@@ -47,6 +48,8 @@ public final class ActorTestProbe<Message: Codable>: @unchecked Sendable {
     internal let internalRef: _ActorRef<ProbeCommands>
     internal let exposedRef: _ActorRef<Message>
 
+    private lazy var _internal = _TestProbeInternal(actorSystem: self.system)
```
ktoso, line 51: lazy var is not thread safe, so that's another reason I was hoping we can initialize it during the test probe init, rather than like this

Sources/DistributedActorsTestKit/TestProbes.swift
```diff
@@ -767,3 +776,104 @@ extension ActorTestProbe {
...
+    public func watch<Act>(
+        _ watchee: Act,
+        file: String = #filePath, line: UInt = #line
+    ) async -> Act
+        where Act: DistributedActor,
+        Act.ActorSystem == ClusterSystem
+    {
+        _ = await self._internal.whenLocal { __secretlyKnownToBeLocal in // TODO(distributed): this is annoying, we must track "known to be local" in typesystem instead
+            __secretlyKnownToBeLocal.watchTermination(of: watchee, file: file, line: line)
+        }
```
ktoso, line 804: ok that's fine 👍

Sources/DistributedActorsTestKit/TestProbes.swift
```diff
@@ -767,3 +776,104 @@ extension ActorTestProbe {
...
+        return watchee
+    }
+
+    /// Checks that the actor has been terminated.
+    ///
+    /// - Warning: Remember to first `watch` the actor you are expecting termination for,
+    ///            otherwise the termination signal will never be received.
+    ///
+    /// - SeeAlso: `DeathWatch`
+    // TODO: expectTermination(of: ...) maybe nicer wording?
```
ktoso, line 814: either works for me, whichever you prefer :)

ktoso: Maybe you're right actually -- Terminated wording is from "old world" where it literarily was "the Terminated" message

Sources/DistributedActorsTestKit/TestProbes.swift
```diff
@@ -767,3 +776,104 @@ extension ActorTestProbe {
...
+        where Act: DistributedActor,
+        Act.ActorSystem == ClusterSystem
+    {
+        _ = await self._internal.whenLocal { __secretlyKnownToBeLocal in // TODO(distributed): this is annoying, we must track "known to be local" in typesystem instead
+            __secretlyKnownToBeLocal.watchTermination(of: watchee, file: file, line: line)
+        }
+        return watchee
+    }
+
+    /// Checks that the actor has been terminated.
```
ktoso, line 808:
```suggestion
    /// Suspends waiting for the actor to become terminated.
```

Sources/DistributedActorsTestKit/TestProbes.swift
yim-lee: I made it `lazy` because I noticed it gets created even when not needed and that messes up the traversal tests. Maybe one option is to add a flag to the initializer and make `_internal` optional?

ktoso: Ok I see, that's the opposite way around we need to be fixing though -- the traversal tests don't matter I'll check which ones :)

ktoso: I.e. this should be eager, especially since the current take is not entirely threadsafe and while it should nto matter too often in tests accessing the probe they must be threadsafe to access from anywhere :)

yim-lee: ok, i will delete the traversal test and always initialize this in `init`.

ktoso (approved):
LGTM! We simplified the timeout a bit too :)

Sources/DistributedActorsTestKit/TestProbes.swift
yim-lee: actually, the traversal test passes with `private let _internal: _TestProbeInternal`, so i will keep the test for now.

Sources/DistributedActorsTestKit/TestProbes.swift
yim-lee: not storing anymore. just use it in `init`.

yim-lee:
Test failures: https://github.com/apple/swift-distributed-actors/issues/945, https://github.com/apple/swift-distributed-actors/issues/992 (new)

Sources/DistributedActorsTestKit/TestProbes.swift
ktoso: ok :)

ktoso:
Fixed the timeout task issue: https://github.com/apple/swift-distributed-actors/pull/991/commits/19bbec083b9804efcfc3e5019f43b42a2a82357d

Was a bit brain bending but I guess makes sense... the cancellation could have executed synchronously and trigger the cancellation to complete the `select` which we absolutely did not want

ktoso:

ktoso:
Ok I really need to stabilize `ClusterLeaderActionsClusteredTests.test_singleLeader`

ktoso:
otherwise all good

PR #995, opened 2022-07-08
Add .devcontainer

PopFlamingo:
[Development containers](https://containers.dev) are a great way of defining development environments that exactly match the project requirements. They have native integration with Visual Studio Code and are also used by Github Codespaces to create their development environments with a single click right from GitHub.

### Motivation:

Make contributing easier by avoiding manual dev container configuration and enabling easier usage of features such as Codespaces. Contributors can be sure the environment they get automatically is always the right one.

### Modifications:

Adds a .devcontainer definition based on the `swiftlang/swift:nightly-5.7-focal` Docker image.

### Other notes

While this PR doesn't provide it and only focuses on just providing a simple dev environment, it could also be possible to define suggestions to install certain extensions, such as the official Swift Visual Studio Code extension.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,1 @@
+FROM swiftlang/swift:nightly-5.7-focal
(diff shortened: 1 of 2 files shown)


ktoso:
Hm, not sure we can... need to double check.

ktoso:
Thanks for the PR, seems we can use these :-)

PR #996, opened 2022-07-09
Silence OpLogDistributedReceptionist.sendAckOps timeout logs

### LINKED ISSUE #993: Silence remote call timeout logging
Screen Shot 2022-07-08 at 18 29 43

We should silence timeout on remote call logging, this is not actionable or interesting to users. Healthchecks will be timing out sometimes and it is expected.

PopFlamingo:
I can't really easily reproduce the specific error that caused the logging, but as far as I can tell this is where it came from and I also didn't see this message anymore in my logs.

I was wondering if I should just drop it entirely or maybe still log it under the trace level for instance?

- Resolves #993

### DIFF
--- a/Sources/DistributedActors/Cluster/Reception/OperationLogDistributedReceptionist.swift
+++ b/Sources/DistributedActors/Cluster/Reception/OperationLogDistributedReceptionist.swift
@@ -629,11 +629,9 @@ extension OpLogDistributedReceptionist {
                 try await peerReceptionistRef.ackOps(until: latestAppliedSeqNrFromPeer, by: self)
             } catch {
                 switch error {
-                case RemoteCallError.clusterAlreadyShutDown:
+                case RemoteCallError.clusterAlreadyShutDown, is TimeoutError:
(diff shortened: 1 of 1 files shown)


ktoso:
Ah this looks right, thank you -- would you mind merging this `case` with the one above it? It's the same intent between them (so that the handling has a comment)

PopFlamingo:
@ktoso Done! Should I also make the comment more generic?

ktoso:
Hm yeah, just that timeouts are an expected thing here really

PopFlamingo:
Ok, I modified it a bit.

ktoso (approved):

ktoso:
LGTM thank you!

ktoso:
We have one very flaky test since recently, I'll get to it soon. Tests good other than that one

PopFlamingo:
Oops it looks like I missed something and the error was actually ` RemoteCallError.timedOut` and not `TimeoutError`, will make another PR to fix it.

PR #997, opened 2022-07-09
Use correct error type for timeout error silencing

PopFlamingo:
Fix silenced error type for OpLogDistributedReceptionist.sendAckOps

### Motivation:

The error type the was silenced by #996 was actually wrong, this fixes it.


PopFlamingo:
Sorry, I should have double checked.

ktoso:
Thanks :)

PR #1003, opened 2022-07-13
Fix logging level setting issue

### LINKED ISSUE #1002: Settings log level isn't respected
When setting a specific log level with the cluster system settings, they aren't taken into account.
For instance when using:
```swift
settings.logging.logLevel = .critical
```

I can still see `info` messages being logged.

PopFlamingo:
Avoid the logging level settings from being overridden when replacing the "initialization logger" with a new one.

### Motivation:

The logging settings were overridden at some point and it caused the settings to not be taken into account (see #1002)

### Modifications:

Copy the logging level that had been set on the initial logger to apply it again to the logger that replaces it.

### Result:

- Resolves #1002


ktoso:
Nice catch, thanks 👍

ktoso (approved):

PR #1006, opened 2022-07-15
Make installSwiftBacktrace public before we remove it

ktoso:
This allows people to disable it; 

We'll want to remove the dependency in general though - and only enable it in tests for example.


ktoso:

Issue #999, opened 2022-07-11
[api evolution] Make all public enums into structs with static properties


ktoso:
We need to do the usual trick of 

```swift
public struct Thing { 
  let underlying: _Thing

  internal enum _Thing { 
  case example
  }

  public static example = .init(.example)
 
  internal init(_ underlying: _Thing) { 
    self.underlying = underlying
  }
}
```

to ensure future API compatibility -- for example for all settings, like singleton allocation strategy etc.

ktoso:
This is required for 1.0

ktoso:
For errors the dance is even more involved:

```swift
public struct SomeError: Error, Sendable { 
  let underlying: _Storage

  internal class _Storage { 
    let error: _SomeError
    let file: String
    let line: UInt
  }

  internal enum _SomeError { 
  case something(...)
  }

  internal init(error: _SomeError, file: String = #fileID (or path? I forgot?), line: UInt = #line) { 
    self.underlying = .init(error: error, file: file, line: line)
  }

  var description: String { 
    "(Self.self)(\(error: \(underlying.error), at: \(underlying.file):\(underlying.line))
}
```

or something like that...

Did I get that right @fabianfett @glbrntt ?

ktoso:
Thanks for double checking :)

PR #1023, opened 2022-07-22
Hide VersionVector

yim-lee:
Part of https://github.com/apple/swift-distributed-actors/issues/999


ktoso:
Sounds good! we don't need it public 👍

ktoso (approved):

yim-lee:
I considered moving `VersionVector.proto` to internal too but it would impact `Membership` and `ClusterEvents` so I left it.

ktoso:
Hmmm... makin it internal might have been nice but it is _Proto prefixed so no big deal 👍

PR #1025, opened 2022-07-22
Make public enums into structs or hide them (part 1)

yim-lee:
Part of https://github.com/apple/swift-distributed-actors/issues/999

### DIFF
--- a/Sources/DistributedActors/Cluster/Cluster+Event.swift
+++ b/Sources/DistributedActors/Cluster/Cluster+Event.swift
@@ -24,6 +24,7 @@ extension Cluster {
         case membershipChange(MembershipChange)
         case reachabilityChange(ReachabilityChange)
         case leadershipChange(LeadershipChange)
+        case _PLEASE_DO_NOT_EXHAUSTIVELY_MATCH_THIS_ENUM_NEW_CASES_MIGHT_BE_ADDED_IN_THE_FUTURE
     }
 }
 
--- a/Sources/DistributedActors/Cluster/Downing/DowningSettings.swift
+++ b/Sources/DistributedActors/Cluster/Downing/DowningSettings.swift
@@ -17,35 +17,51 @@ import Logging
 // ==== ----------------------------------------------------------------------------------------------------------------
 // MARK: DowningStrategySettings
 
-public enum DowningStrategySettings {
-    case none
-    case timeout(TimeoutBasedDowningStrategySettings)
+public struct DowningStrategySettings {
+    private enum _DowningStrategySettings {
+        case none
+        case timeout(TimeoutBasedDowningStrategySettings)
+    }
+
+    private let underlying: _DowningStrategySettings
+
+    private init(_ underlying: _DowningStrategySettings) {
+        self.underlying = underlying
+    }
 
     func make(_ clusterSystemSettings: ClusterSystemSettings) -> DowningStrategy? {
-        switch self {
+        switch self.underlying {
         case .none:
             return nil
         case .timeout(let settings):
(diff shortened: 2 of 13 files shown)


Sources/DistributedActors/Cluster/Cluster+Event.swift
```diff
@@ -19,11 +19,35 @@ extension Cluster {
...
+            .init(.membershipChange(membershipChange))
+        }
+
+        public static func reachabilityChange(_ reachabilityChange: ReachabilityChange) -> Event {
+            .init(.reachabilityChange(reachabilityChange))
+        }
+
+        public static func leadershipChange(_ leadershipChange: LeadershipChange) -> Event {
+            .init(.leadershipChange(leadershipChange))
+        }
```
ktoso, line 50: Hmmm... so this one I'd kind of want to keep as an enum to be honest :\

The primary way to use it is to switch over it in an async for loop of events...

Perhaps we should do this `case _PLEASE_DO_NOT_EXHAUSTIVELY_MATCH_THIS_ENUM_NEW_CASES_MIGHT_BE_ADDED_IN_THE_FUTURE` trick for this one specifically... other SSWG libs do this for such cases.

Sources/DistributedActors/Cluster/Reception/OperationLog.swift
```diff
@@ -151,12 +151,12 @@ extension OpLog.SequencedOp: Codable where Op: Codable {
 extension OpLog.SequencedOp: Equatable where Op: Equatable {}
 
 extension OpLog.SequencedOp.SequenceRange: Codable {
-    public enum DiscriminatorKeys: String, Codable {
+    enum DiscriminatorKeys: String, Codable {
```
ktoso, line 154: yep good to keep internal!

Sources/DistributedActors/Cluster/Downing/DowningSettings.swift
```diff
@@ -78,4 +94,16 @@ public enum OnDownActionStrategySettings {
...
+    ///
+    /// When using this mode you should take special care to implement some form of shutting down of this node (!).
+    /// As a ``Cluster/MemberStatus/down`` node is effectively useless for the rest of the cluster -- i.e. other
+    /// members MUST refuse communication with this down node.
+    public static let none: OnDownActionStrategySettings = .init(.none)
+
+    /// Upon noticing that this member is marked as [.down], initiate a shutdown.
+    public static func gracefulShutdown(delay: Duration) -> OnDownActionStrategySettings {
+        .init(.gracefulShutdown(delay: delay))
+    }
```
ktoso, line 108: good one, thank you -- yes this is a good candidate for this treatment, LGTM

ktoso:
Downing and OpLog LGTM, the Event I guess we may need to do the terrible approach instead... :-/
One day we'll get better enums in swift, like the stdlib has... but until then we have the hack heh :\

Sources/DistributedActors/Cluster/Cluster+Event.swift
yim-lee: amended https://github.com/apple/swift-distributed-actors/pull/1025/commits/7ea7b1f46316ddbaff5d6861cbd6799bd9b80e02

Internally we just log error and/or ignore the "fake" event. Is that ok? I don't think we should do `default` because we will forget to update the `switch` when we add new event in the future. It doesn't feel right to throw error or crash either.

ktoso: Agreed, that looks good

ktoso:
Flaky https://github.com/apple/swift-distributed-actors/issues/971

PR #1026, opened 2022-07-25
Avoid errors in old toolchains, add to-be-removed convenience keyword

ktoso:
It was reported that beta-2 toolchain errors out on the missing `convenience` in this init. 

This is actually fixed by https://github.com/apple/swift-evolution/blob/main/proposals/0327-actor-initializers.md which @kavon implemented, so as soon as we get a stable Swift which does not need the convenience keyword we can drop it.

Adding for now to make life of early adopters easier.


ktoso:
This will produce:

```
/Users/ktoso/code/swift-distributed-actors/Sources/DistributedActors/Pattern/WorkerPool.swift:87:24: warning: initializers in actors are not marked with 'convenience'; this is an error in Swift 6
    public convenience init(selector: Selector, actorSystem: ActorSystem) async throws {
```

but let's be ok with that until we have stable 5.7, and then we flip over and remove all those `convenience`

PR #1032, opened 2022-07-26
=receptionist lookup() trying to cast stub actor

ktoso:
Receptionist would bug out when trying to lookup() due to a wrongly implemented cast.

I'll want to enable all the other tests here and spend more time hardening for our next milestone.


ktoso:
Resolves rdar://97586760

Sources/DistributedActors/Cluster/Reception/OperationLogDistributedReceptionist.swift
```diff
@@ -391,7 +391,7 @@ extension OpLogDistributedReceptionist: LifecycleWatch {
 
         // self.instrumentation.listingPublished(key: message._key, subscribers: 1, registrations: registrations.count) // TODO(distributed): make the instrumentation calls compatible with distributed actor based types
         let guests = Set(registrations.compactMap { versioned in
-            versioned.actor.underlying as? Guest
+             try? Guest.resolve(id: versioned.actor.underlying.id as! ActorID, using: self.actorSystem)
```
ktoso, line 394: I'll try to see if we can store the right one right away instead.

Sources/DistributedActors/DistributedActor+Internal.swift
```diff
@@ -68,10 +68,7 @@ extension DistributedActor where ActorSystem == ClusterSystem {
     }
 }
 
-distributed actor StubDistributedActor {
+internal distributed actor StubDistributedActor {
     typealias ID = ClusterSystem.ActorID
     typealias ActorSystem = ClusterSystem
-
-    // TODO: this is just to prevent a DI crash because of enums without cases and Codable
-    distributed func _noop() {}
```
ktoso, line 76: wohoo, compiler no more crashing on empty DA 🥳

ktoso:
Sigh always one flaky test somewhere.

I'll spend this week tracking them all down 💯

yim-lee (approved):

Issue #1034, opened 2022-07-26
FAILED: ClusterSingletonPluginClusteredTests.test_remoteCallShouldFailAfterAllocationTimedOut


ktoso:
```
20:28:19 Test Case 'ClusterSingletonPluginClusteredTests.test_remoteCallShouldFailAfterAllocationTimedOut' started at 2022-07-26 11:28:07.389
20:28:19 <EXPR>:0: error: ClusterSingletonPluginClusteredTests.test_remoteCallShouldFailAfterAllocationTimedOut : threw error "
20:28:19         try await self.assertMemberStatus(on: second, node: firstNode, is: .down, within: .seconds(10))
20:28:19                              ^~~~~~
20:28:19 error: MembershipError(awaitStatusTimedOut(10.0 seconds, Optional(MembershipError(statusRequirementNotMet(expected: DistributedActors.Cluster.MemberStatus.down, found: Member(sact://first:11906826780297485806@127.0.0.1:7111, status: joining, reachability: unreachable)), at: DistributedActors/ClusterControl.swift:249))), at: DistributedActors/ClusterControl.swift:304)"
```

total:

```

20:28:19 <EXPR>:0: error: ClusterSingletonPluginClusteredTests.test_remoteCallShouldFailAfterAllocationTimedOut : threw error "
20:28:19         try await self.assertMemberStatus(on: second, node: firstNode, is: .down, within: .seconds(10))
20:28:19                              ^~~~~~
20:28:19 error: MembershipError(awaitStatusTimedOut(10.0 seconds, Optional(MembershipError(statusRequirementNotMet(expected: DistributedActors.Cluster.MemberStatus.down, found: Member(sact://first:11906826780297485806@127.0.0.1:7111, status: joining, reachability: unreachable)), at: DistributedActors/ClusterControl.swift:249))), at: DistributedActors/ClusterControl.swift:304)"
20:28:19 ------------------------------------- ClusterSystem(first, sact://first@127.0.0.1:7111) ------------------------------------------------
20:28:19 Captured log [first][2022-07-26 11:28:07.3930] [ClusterSystem.swift:979] [trace] Assign identity
20:28:19 // metadata:
20:28:19 // "actor/id": /system/clustEventStream
20:28:19 // "actor/type": ClusterEventStreamActor
20:28:19 Captured log [first][2022-07-26 11:28:07.3930] [ClusterSystem.swift:992] [trace] Actor ready
20:28:19 // metadata:
[...]

PR #1059, opened 2022-08-30
=multi-node improve printouts to always be one-by-one (logs)

ktoso:
New receptionist was too aggressive in what it deemed acceptable to be send onNext to the reception stream.

Concurrent messages are fine; specifically if [first:1] then [first:1, second:1] happens, these may have been concurrent in vector clock meaning, but it means that the event has more information so we should be emitting it.

### DIFF
--- a/Sources/DistributedActors/Cluster/Reception/OperationLogDistributedReceptionist.swift
+++ b/Sources/DistributedActors/Cluster/Reception/OperationLogDistributedReceptionist.swift
@@ -295,6 +295,22 @@ extension OpLogDistributedReceptionist: LifecycleWatch {
         }
     }
 
+    public nonisolated func checkIn<Guest>(
+        _ guest: Guest
+    ) async where Guest: DistributedActor, Guest.ActorSystem == ClusterSystem {
+        guard let keyID: String = guest.id.metadata.receptionID else {
+            fatalError("""
+                       Attempted to \(#function) distributed actor without `@ActorID.Metadata(\\.receptionKey)` set on ActorID!
+                       Please set the metadata during actor initialization.
+                       """)
+        }
+        let key = DistributedReception.Key(Guest.self, id: keyID)
+
+        await self.whenLocal { myself in
+            await myself._checkIn(guest, with: key)
+        }
+    }
+
     // 'local' implementation of checkIn
     private func _checkIn<Guest>(
         _ guest: Guest,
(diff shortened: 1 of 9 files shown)


ktoso:
Joining timeout in 

```
19:32:33 Test Case 'ClusterSingletonPluginClusteredTests.test_remoteCallShouldFailAfterAllocationTimedOut' started at 2022-08-30 10:32:20.538
19:32:33 <EXPR>:0: error: ClusterSingletonPluginClusteredTests.test_remoteCallShouldFailAfterAllocationTimedOut : threw error "
19:32:33         try await self.assertMemberStatus(on: second, node: firstNode, is: .down, within: .seconds(10))
19:32:33                              ^~~~~~
19:32:33 error: MembershipError(awaitStatusTimedOut(10.0 seconds, Optional(MembershipError(statusRequirementNotMet(expected: DistributedActors.Cluster.MemberStatus.down, found: Member(sact://first:1722502985102675705@127.0.0.1:7111, status: joining, reachability: unreachable)), at: DistributedActors/ClusterControl.swift:252))), at: DistributedActors/ClusterControl.swift:338)"
```

ktoso:
Seems to be same as https://github.com/apple/swift-distributed-actors/issues/1034 - the nodes seem to sometimes need longer to join up for the test.

ktoso:
Validation test is green which is good, this one would have triggered the bug this is fixing :-)

Sources/DistributedActors/Cluster/Reception/OperationLogDistributedReceptionist.swift
```diff
@@ -295,6 +295,22 @@ extension OpLogDistributedReceptionist: LifecycleWatch {
...
+    ) async where Guest: DistributedActor, Guest.ActorSystem == ClusterSystem {
+        guard let keyID: String = guest.id.metadata.receptionID else {
+            fatalError("""
+            Attempted to \(#function) distributed actor without `@ActorID.Metadata(\\.receptionKey)` set on ActorID!
+            Please set the metadata during actor initialization.
+            """)
+        }
+        let key = DistributedReception.Key(Guest.self, id: keyID)
+
+        await self.whenLocal { myself in
```
yim-lee, line 309: This reads very nicely 👍

Sources/DistributedActors/Cluster/Reception/OperationLogDistributedReceptionist.swift
```diff
@@ -295,6 +295,22 @@ extension OpLogDistributedReceptionist: LifecycleWatch {
         }
     }
 
+    public nonisolated func checkIn<Guest>(
+        _ guest: Guest
+    ) async where Guest: DistributedActor, Guest.ActorSystem == ClusterSystem {
+        guard let keyID: String = guest.id.metadata.receptionID else {
+            fatalError("""
+            Attempted to \(#function) distributed actor without `@ActorID.Metadata(\\.receptionKey)` set on ActorID!
```
yim-lee, line 303:
```suggestion
            Attempted to \(#function) distributed actor without `@ActorID.Metadata(\\.receptionID)` set on ActorID!
```

Sources/DistributedActors/Receptionist/DistributedReceptionist.swift
```diff
@@ -429,22 +434,18 @@ internal final class AnyDistributedReceptionListingSubscription: Hashable, @unch
             // the seen vector was unaffected by the merge, which means that the
             // incoming registration version was already seen, and thus we don't need to emit it again
             return false
-        case .happenedAfter:
+        case .happenedAfter, .concurrent:
```
yim-lee, line 437: Ah, makes sense I think.

yim-lee (approved):
One typo but otherwise LGTM 👍

Sources/DistributedActors/Cluster/Reception/OperationLogDistributedReceptionist.swift
ktoso: I recently realized we don't always have to do the `__` trickery by the way :)

Sources/DistributedActors/Cluster/Reception/OperationLogDistributedReceptionist.swift
ktoso: whoops thanks!

PR #1062, opened 2022-09-20
Add Swift 5.7 support to legacy release/0.x branch.

mrstegeman:
This adds support for Swift 5.7, but only to the legacy version.

### Motivation:

Some projects are still using the legacy version but would like to use Swift 5.7. The dependencies of this project require updating for each Swift version.

### Modifications:

* Updated dependencies for Swift 5.7.
* Worked around an issue in swift-metrics, but only did so for Swift 5.7 to keep backwards compatibility.

### Result:

Swift 5.7 is now supported.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -296,8 +296,6 @@ var dependencies: [Package.Dependency] = [
     // ~~~ SSWG APIs ~~~
 
     .package(url: "https://github.com/apple/swift-log.git", from: "1.0.0"),
-    // swift-metrics 1.x and 2.x are almost API compatible, so most clients should use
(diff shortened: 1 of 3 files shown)


ktoso:

ktoso (approved):
LGTM, we'll also setup a 5.7 CI for here I suppose.

Thank you!

mrstegeman:
Thanks! Hopefully won't need it much longer. A new tag/release would be great after you merge, if you don't mind.

ktoso:
Yup, I'll cut a release right away for y'all :)

Issue #1064, opened 2022-10-05
Fix build, consuming from other packages


ktoso:
```
'InternalPlugins/MultiNodeTest' is not a valid path for path-based dependencies; use relative or absolute path instead. in https://github.com/apple/swift-distributed-actors
```

when consumed from another package. We must fix this before the next tag.

PR #1068, opened 2022-10-17
=build fix how we declare the multi-node plugin

ktoso:
Actual solution for https://github.com/apple/swift-distributed-actors/issues/1064

This stops making use of `InternalPlugins` so it builds correctly again when depended on I think.
I'll verify in the morning though.

Resolves https://github.com/apple/swift-distributed-actors/issues/1064


ktoso:

ktoso:
Ok confirmed this works (downstreams won't fail with the plugin but also don't yet it, good)

PR #1077, opened 2022-10-27
!all node renames: node -> endpoint; unique node -> node

ktoso:
This is a massive rename, but an important one.

Discussed this with NIO team a while ago and we arrived at this conclusion.

The `Node` and `UniqueNode` have been confusing for newcomers for a long time.
Instead we offer the following:

- old name `Node` is now **`Cluster.Endpoint`** - the purpose of this type is to be protocol + host + port, BEFORE we know the unique identity of a node. It is used to "connect over there please" or "regardless of node identity, that one" work with nodes. It is used mostly in `join(endpoint:)` as well as configuring where we should `bind` on start of the cluster.
- old name `UniqueNode` is now **`Cluster.Node`** - which makes more sense as this is the actual node identity; A node of course is unique, so the previous wording may have been confusing.

This also makes sense since progression of a "node" while joining a cluster is: 
- `Cluster.Endpoint` - we don't know it's identity, or if it even has a running cluster node at all.
    -  `Endpoint = Protocol + (Name?) + Host + Port`
- → `Cluster.Node` - we know it exists and have its unique ID. We are connecting to it, but it may REJECT our join attempt, e.g. if we're version incompatible etc. 
    - `Node = Endpoint + Unique Node ID`
- → `Cluster.Member` - we have successfully connected and formed a cluster with that Node. From now on we also know it's membership status and reachability.
    - `Member = Node + Membership information`


Sources/DistributedCluster/Cluster/ClusterEventStream.swift
```diff
@@ -128,7 +128,7 @@ internal distributed actor ClusterEventStreamActor: LifecycleWatch {
 
     static var props: _Props {
         var ps = _Props()
-        ps._knownActorName = "clustEventStream"
+        ps._knownActorName = "clusterEventStream"
```
yim-lee, line 131: 😱

yim-lee (approved):
👍 (didn't look at tests)

Sources/DistributedCluster/Cluster/ClusterEventStream.swift
ktoso: haha yeah lol

Sources/DistributedActorsTestKit/Cluster/ClusteredActorSystemsXCTestCase.swift
```diff
@@ -201,36 +201,36 @@ open class ClusteredActorSystemsXCTestCase: XCTestCase {
         ensureWithin: Duration? = nil, ensureMembers maybeExpectedStatus: Cluster.MemberStatus? = nil,
         file: StaticString = #filePath, line: UInt = #line
     ) async throws {
-        node.cluster.join(node: other.cluster.uniqueNode.node)
+        node.cluster.join(endpoint: other.cluster.node.endpoint)
```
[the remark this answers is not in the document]
ktoso: yeah quite some renames here... but all got done <3

ktoso:
Renames went too far -- the test must grep for `Node already leaving...` should be good now.

ktoso:
5.7 failure on a known flaky `InterceptorTests.test_interceptor_shouldInterceptSignals` which actually we should just remove -- interceptors are not a thing anymore.

other green so I'll merge and remove the InterceptorTests.test_interceptor_shouldInterceptSignals https://github.com/apple/swift-distributed-actors/issues/970 as we said in there alreyad

PR #1088, opened 2022-11-09
Use Serialization instead of JSONEncoder in recordArgument()

orobio:
Int arguments were serialized as Strings, but the deserializer expected bytes. I found these commented lines of code that seemed to be the correct way of serializing instead of using JSONEncoder, but I'm not sure if that is true. If it is, I guess the reason these were comments might still apply, but it does work in my simple application and all tests are successful.


ktoso:

ktoso:

ktoso:

Issue #1108, opened 2023-02-16
Main broken due to Task.select removal from async-algorithms


ktoso:
https://github.com/apple/swift-async-algorithms/pull/217/

We need to remove our use of Task.select.

PR #1109, opened 2023-02-17
-testkit remove the Task.select based timeout impl, make a group based one

ktoso:
We were depending on a work in progress async-algorithms and Task.select was removed without replacement.

We only used this in testing though, so I reimplemented it as a task group based one.
It's not quite the same, but consensus is that such impl is "more correct"

resolves https://github.com/apple/swift-distributed-actors/issues/1108


ktoso:
5.8 CI is not set up, this will let us confirm if 5.8 is passing in the other PR.

ktoso:
This is to unblock you @weissi

PR #1112, opened 2023-03-01
actor shell slimmed down a bit, adjust test

ktoso:
### Motivation:

Test fails on macos; Test should be adjusted to reflect smaller shell size

### Modifications:

Adjust shell instance size (down)

### Result:

passing tests on macos


ktoso:
Flaky test: ClusterLeaderActionsClusteredTests.test_up_ensureAllSubscribersGetMovingUpEvents https://github.com/apple/swift-distributed-actors/issues/971

PR #1044, opened 2022-07-29
[WIP] async leadership WIP

ktoso:
Eventually resolves https://github.com/apple/swift-distributed-actors/issues/1040

Very WIP, need to switch over to another task for now.

PR #1117, opened 2023-04-24
[Leadership] Enable implementing leadership decisions outside core lib

ktoso:
### Motivation:

It was not possible to implement a "real" leadership election using whatever mechanism one could dream up using the current exposed methods.

We now expose `assumeLeader` which makes the local assumption change, and therefore it is possible to call that method from _real_ leadership election implementations. We should offer more of those AND document how to implement your custom one -- the example impl could be a bully implementation for example.

The test uses a dumb implementation that is only there for showcasing how to use the receptionist to discover peers in the process.

### Modifications:

- `LeaderElection` -> `LegacyLeaderElection`, because it was using NIO Futures
- add `cluster.assumeLeader`
- add a few methods to await `waitToBecomeLeader` and `waitForLeader(atLeast:within:)`

### Result:

Ability to implement ANY leadership election mechanism and inform the cluster node which node should be considered the leader.

The leader decision is NOT gossiped. This is entirely up to the leader election mechanism to propagate this information.

- Replaces https://github.com/apple/swift-distributed-actors/pull/1044


ktoso:

PR #1121, opened 2023-06-06
Add helper method to check if a plugin is installed

yim-lee:
Motivation:
`PluginsSettings` doesn't allow same plugin to be installed more than once, but there is no way to check if a plugin is installed to avoid duplicate installation calls.

Modification:
Add `inInstalled` to `PluginsSettings`.


ktoso (approved):
LGTM thanks! Sounds good to allow checking.

I was thinking if the install should just check but this is better since the install may have some specific configuration I guess.

PR #1122, opened 2023-06-29
ClusterControl.joined should cause joining; fix waiting issues

ktoso:
bug: We wrongly waited on the `self.node` to become up but should wait for the other node in `joined(node:)`

improvement: We should initiate joining when not already trying to and someone asks `try await joined(node:)` since the equivalent `join()` makes it seem like joined would do this as well

### DIFF
--- a/Sources/DistributedCluster/Cluster/ClusterControl.swift
+++ b/Sources/DistributedCluster/Cluster/ClusterControl.swift
@@ -194,8 +194,9 @@ public struct ClusterControl {
     public func joined(node: Cluster.Node, within: Duration) async throws -> Cluster.Member {
         // waiting to be "joined" means having passed the "joining" member status
         if let member = await self.membershipSnapshot.member(node),
-           member.status > .joining {
-                return member
+           member.status > .joining
+        {
+            return member
         }
 
         self.join(node: node) // kick off the joining process in case we're not trying to already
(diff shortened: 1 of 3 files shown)


Sources/DistributedCluster/Cluster/ClusterControl.swift
```diff
@@ -175,19 +192,33 @@ public struct ClusterControl {
...
     ///
     /// - Parameters
-    ///   - node: The node to be joined by this system.
+    ///   - endpoint: The endpoint to be joined by this system.
     ///   - within: Duration to wait for.
     ///
     /// - Returns `Cluster.Member` for the joined node.
     @discardableResult
     public func joined(endpoint: Cluster.Endpoint, within: Duration) async throws -> Cluster.Member? {
-        try await self.waitFor(self.node, .up, within: within)
```
ktoso, line 190: actual bug here, we must wait for `endpoint` not `self.node` ;)

Sources/DistributedCluster/Cluster/ClusterControl.swift
```diff
@@ -314,7 +347,7 @@ public struct ClusterControl {
                 throw Cluster.MembershipError(.notFound(node, in: membership))
             }
 
-            if atLeastStatus <= foundMember.status {
```
ktoso, line 317: another mini bug; we should error out if the expected status is lower than the current one.

if they are the same, don't error -- we succeeded; the test covers this now

ktoso:
The failure is `ClusterSingletonPluginClusteredTests.test_singletonByClusterLeadership_withLeaderChange` and I've actually been looking into it now, may solve. It's a bit tricky coming back to this after a while, but I'll figure it out -- it's an unexpected dead letter :)

ktoso:
Failed: https://github.com/apple/swift-distributed-actors/issues/1124 needs investigation

ktoso:
This improves things but we definitely still have some issues to look into here, I'll schedule time to work on those consistently now.

yim-lee (approved):

PR #1126, opened 2023-07-03
ClusterControl.joined should cause joining for simpler usage

ktoso:
This hardens singleton handover, sadly it also hits a bug in IRGen which made the impl uglier than I would have wanted to.

This is not to be reviewed yet, I want to give it some runs on CI

### DIFF
--- a/Sources/DistributedCluster/Backoff.swift
+++ b/Sources/DistributedCluster/Backoff.swift
@@ -146,8 +146,6 @@ public struct ConstantBackoffStrategy: BackoffStrategy {
 ///
 /// - SeeAlso: Also used to configure `_SupervisionStrategy`.
 public struct ExponentialBackoffStrategy: BackoffStrategy {
-    // TODO: clock + limit "max total wait time" etc
-
     /// Default values for the backoff parameters.
     public enum Defaults {
         public static let initialInterval: Duration = .milliseconds(200)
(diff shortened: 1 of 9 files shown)


Sources/DistributedCluster/Backoff.swift
```diff
@@ -208,6 +206,39 @@ public struct ExponentialBackoffStrategy: BackoffStrategy {
...
+    /// Attempt to execute the passed `operation` at most `maxAttempts` times while applying the expected backoff strategy.
+    ///
+    /// - Parameters:
+    ///   - operation: The operation to run, potentially multiple times until successful or maxAttempts were made
+    /// - Throws: the last error thrown by `operation`
+    public func attempt<Value>(_ operation: () async throws -> Value) async throws -> Value {
+        var backoff = self
+        var lastError: Error? = nil
+        defer {
+            pprint("RETURNING NOW singleton")
```
yim-lee, line 218: This and L224 probably not intended to keep

Sources/DistributedCluster/DeadLetters.swift
```diff
@@ -220,6 +220,10 @@ public final class DeadLetterOffice {
                 return metadata
             }()
         )
+
+        pinfo("ALL ACTORS: \(self.system?.cluster.node)")
+        self.system?._printTree()
+        print("\n\n======")
```
yim-lee, line 226: intended to keep?

yim-lee (commented):
I _think_ changes in this PR are already covered by https://github.com/apple/swift-distributed-actors/pull/1127 and https://github.com/apple/swift-distributed-actors/pull/1122?

ktoso:
Oh whoops, I did something wrong here with PRs. Let me check if there is anything missing from main still

ktoso:
You're right my bad, this was subsumed already.

PR #1132, opened 2023-07-18
Prevent distributed singletons created during other's init from duping IDs

### LINKED ISSUE #1131: DA spawned by a ClusterSingleton inherits _Props and leads to duplicateActorPath error
I have a cluster singleton (e.g., `Singleton`) that spawns another distributed actor (e.g., `Greeter`) to do some work. 

```swift
distributed actor Singleton: ClusterSingleton {

    private let greeter: Greeter

    init(actorSystem: ActorSystem) {
        self.actorSystem = actorSystem
        self.greeter = Greeter(actorSystem: actorSystem)
    }
}

distributed actor Greeter {

    distributed func greet() {
        print("Hello!")
    }
}
```

I have a test that starts the cluster singleton in a single-node cluster like so:

```swift
        let singleton = try await clusterSystem.singleton.host(name: "test-singleton") { actorSystem in
            Singleton(actorSystem: actorSystem)
        }
```

The test crashes with error:

```
DistributedCluster/ClusterSystem.swift:759: Fatal error: 'try!' expression unexpectedly raised an error: ClusterSystemError(duplicateActorPath(path: /user/test-singleton), at: DistributedCluster/Refs.swift:642)
```

Upon inspecting trace logs I see that `Greeter` has inherited the `_Props` of `Singleton`, which results in the two actors having the same id:

```
2023-07-08T11:31:13-0700 trace test : actor/id=[$wellKnown: test-singleton] actor/type=Singleton cluster/node=sact://test@127.0.0.1:9001 [DistributedCluster] Assign identity
2023-07-08T11:31:13-0700 trace test : actor/id=[$wellKnown: test-singleton] actor/type=Greeter cluster/node=sact://test@127.0.0.1:9001 [DistributedCluster] Assign identity
2023-07-08T11:31:13-0700 trace test : actor/id=[$wellKnown: test-singleton] actor/type=Greeter cluster/node=sact://test@127.0.0.1:9001 [DistributedCluster] Actor ready
2023-07-08T11:31:13-0700 trace test : actor/id=[$wellKnown: test-singleton] actor/type=Singleton cluster/node=sact://test@127.0.0.1:9001 [DistributedCluster] Actor ready
```

The error does NOT occur if `Greeter` is initialized inside a distributed func invoked outside:

```swift
distributed actor Singleton: ClusterSingleton {

    private var greeter: Greeter?

    init(actorSystem: ActorSystem) {
        self.actorSystem = actorSystem
    }
    
    distributed func initGreeter() {
        self.greeter = Greeter(actorSystem: actorSystem)
    }
}

let singleton = try await clusterSystem.singleton.host(name: "test-singleton") { actorSystem in
    Singleton(actorSystem: actorSystem)
}
try await singleton.initGreeter()
```

ktoso:
### Motivation:

We need to "consume" the sneaky way we sneak in the "use this ID" into `assignID`. Without this we'd assign the same ID if someone were to create a singleton DURING the init() of another singleton.

I need to work around a few aggressive assertions that are trying to prevent users doing bad stuff that we actually do want to do here, but this'll work ok.

### Modifications:

We "consume" the "props for spawn" during assignID as it should be. This works because that's code generated as first thing in every distributed actor init.

### Result:

resolves #1131
rdar://112434926

relates to rdar://112433423 which is about a language feature to enable control over this properly.

### DIFF
--- a/Sources/DistributedCluster/Cluster/SWIM/ClusterMembership+Converters.swift
+++ b/Sources/DistributedCluster/Cluster/SWIM/ClusterMembership+Converters.swift
@@ -29,7 +29,11 @@ extension ClusterMembership.Node {
     }
 
     func swimShell(_ system: ClusterSystem) -> SWIMActor {
-        try! SWIMActor.resolve(id: ._swim(on: self.asClusterNode!), using: system) // TODO: the ! is not so nice
+        do {
+            return try SWIMActor.resolve(id: ._swim(on: self.asClusterNode!), using: system)
+        } catch {
+            fatalError("Failed to resolve \(ActorID._swim(on: self.asClusterNode!).detailedDescription): \(error), tree: \n\(system._treeString())")
+        }
     }
 
     var asClusterNode: Cluster.Node? {
--- a/Sources/DistributedCluster/ClusterSystem.swift
+++ b/Sources/DistributedCluster/ClusterSystem.swift
@@ -1013,9 +1013,9 @@ extension ClusterSystem {
             id.metadata.wellKnown = wellKnownName
             id.metadata._props = _PropsShuttle(props: props)
         }
-        if let wellKnownName = props.metadata.remove(forKey: ActorMetadataKeys.__instance.wellKnown) {
-            id.metadata.wellKnown = wellKnownName
-        }
+//        if let wellKnownName = props.metadata.remove(forKey: ActorMetadataKeys.__instance.wellKnown) {
+//            id.metadata.wellKnown = wellKnownName
+//        }
 
         self.log.trace("Assign identity", metadata: [
             "actor/type": "\(actorType)",
--- a/Sources/DistributedCluster/ClusterSystem.swift
+++ b/Sources/DistributedCluster/ClusterSystem.swift
@@ -763,7 +763,7 @@ extension ClusterSystem: _ActorRefFactory {
             return try self._spawn(using: provider, behavior, id: id, props: props)
             // try!-safe, since the naming must have been correct
         } catch {
-            fatalError("THE ID: \(id.fullDescription)\n THE PROPS: \(id.metadata._props?.props)")
+            fatalError("Failed to _spawn DistributedActor, id: \(id.fullDescription); props: \(id.metadata._props?.props), error: \(error)")
         }
     }
 }
--- a/Sources/DistributedCluster/ClusterSystem.swift
+++ b/Sources/DistributedCluster/ClusterSystem.swift
@@ -646,13 +646,12 @@ extension ClusterSystem: _ActorRefFactory {
         try behavior.validateAsInitial()
 
         let incarnation: ActorIncarnation = props._wellKnown ? .wellKnown : .random()
-
         // TODO: lock inside provider, not here
         // FIXME: protect the naming context access and name reservation; add a test
         let id: ActorID = try self.withNamingContext { namingContext in
             let name = naming.makeName(&namingContext)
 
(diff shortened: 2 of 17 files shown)


Sources/DistributedCluster/Cluster/SWIM/ClusterMembership+Converters.swift
```diff
@@ -29,7 +29,11 @@ extension ClusterMembership.Node {
     }
 
     func swimShell(_ system: ClusterSystem) -> SWIMActor {
-        try! SWIMActor.resolve(id: ._swim(on: self.asClusterNode!), using: system) // TODO: the ! is not so nice
+        do {
+            return try SWIMActor.resolve(id: ._swim(on: self.asClusterNode!), using: system)
+        } catch {
+            fatalError("Failed to resolve \(ActorID._swim(on: self.asClusterNode!).detailedDescription): \(error), tree: \n\(system._treeString())")
```
ktoso, line 35: Just improving error message, should not happen in real code but helps debug when we mess up assignID ;)

Sources/DistributedCluster/ClusterSystem.swift
```diff
@@ -1101,9 +1119,9 @@ extension ClusterSystem {
     {
         /// Prepare a distributed actor context base, such that the reserved ID will contain the interceptor in the context.
         let baseContext = DistributedActorContext(lifecycle: nil, remoteCallInterceptor: interceptor)
-        var id = self._assignID(Act.self, baseContext: baseContext)
+         var id = self._assignID(Act.self, baseContext: baseContext)
```
yim-lee, line 1122: nit: extra space

Sources/DistributedCluster/Props.swift
```diff
@@ -79,11 +90,51 @@ public struct _Props: @unchecked Sendable {
...
+        p._wellKnown = self._wellKnown
+        return p
+    }
+
     /// Allows for passing properties to creating a distributed actor.
     @TaskLocal
     internal static var forSpawn: _Props = .init()
 }
 
+/// INTERNAL: Used for shuttling the Props into the ID while performing assignID.
```
yim-lee, line 120: 👍

Tests/DistributedClusterTests/Plugins/ClusterSingleton/ClusterSingletonPluginTests.swift
```diff
@@ -38,4 +39,57 @@ final class ClusterSingletonPluginTests: SingleClusterSystemXCTestCase {
...
+        }
+
+        let singletonID = singleton.id
+        let greeterID = try await singleton.getGreeter().id
+
+        pinfo("singleton proxy    id: \(singletonID)")
+        pinfo("singleton actual   id: \(try await singleton.actualID())")
+        pinfo("singleton(greeter) id: \(greeterID)")
+
+//        singletonID.detailedDescription.shouldContain("test-singleton")
```
yim-lee, line 59: remove?

yim-lee (approved):

Tests/DistributedClusterTests/Plugins/ClusterSingleton/ClusterSingletonPluginTests.swift
ktoso: yup thx

Sources/DistributedCluster/ClusterSystem.swift
ktoso: thx, missed formatter run, fixing

ktoso:
Just one of the jobs had a timeout https://github.com/apple/swift-distributed-actors/issues/1133 failure

PR #1135, opened 2023-08-01
=lifecycle missing synchronization; those methods may be invoked from outside actor

ktoso:
These methods may be invoked from outside the actor, so synchronize them properly as the rest of the container.

I need to review some more code for safety and we should get a tsan build going here, we missed this for quite a while huh.

rdar://112815057

### DIFF
--- a/Sources/DistributedCluster/LifecycleMonitoring/LifecycleWatchContainer.swift
+++ b/Sources/DistributedCluster/LifecycleMonitoring/LifecycleWatchContainer.swift
@@ -198,20 +198,24 @@ extension LifecycleWatchContainer {
 
     /// Performs cleanup of references to the dead actor.
     internal func receiveTerminated(_ terminated: _Signals.Terminated) {
-        self.receiveTerminated(terminated.id)
+        self.receiveTerminated0(terminated.id)
     }
 
-    internal func receiveTerminated(_ terminatedIdentity: ClusterSystem.ActorID) {
+    internal func receiveTerminated(_ terminatedID: ClusterSystem.ActorID) {
         self._lock.wait()
         defer {
             _lock.signal()
         }
 
+        self.receiveTerminated0(terminatedID)
+    }
+
+    private func receiveTerminated0(_ terminatedID: ClusterSystem.ActorID) {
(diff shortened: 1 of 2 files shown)


Sources/DistributedCluster/LifecycleMonitoring/LifecycleWatchContainer.swift
```diff
@@ -50,10 +50,10 @@ final class LifecycleWatchContainer {
...
-//        defer {
-//            _lock.signal()
-//        }
+        self._lock.wait()
+        defer {
+            _lock.signal()
+        }
 
         traceLog_DeathWatch("Clear LifecycleWatchContainer owned by \(self.watcherID)")
         self.watching = [:]
```
yim-lee, line 59: unrelated: is the `FIXME` on L61 still relevant?

yim-lee (approved):
Thanks!

Sources/DistributedCluster/LifecycleMonitoring/LifecycleWatchContainer.swift
```diff
@@ -218,16 +238,6 @@ extension LifecycleWatchContainer {
                 continue
             }
```
yim-lee, line 240: I think we need locking in this method too, for `self.watching` on L216/236.

Sources/DistributedCluster/LifecycleMonitoring/LifecycleWatchContainer.swift
ktoso: I should probably revisit it right away; I think my suspicion there was about if we needed to remove the self watcher ID as well but didn't -- I'll check this during this PR

Sources/DistributedCluster/LifecycleMonitoring/LifecycleWatchContainer.swift
ktoso: Right yeah and a few more, I'm cleaning up this file some more still. thank you for review!

ktoso:
One failure, flaky test need to revisit https://github.com/apple/swift-distributed-actors/issues/1034

PR #1137, opened 2023-09-05
Fix Scheduler implementation for ILP32 platforms such as watchOS

christopherweems:
Fix Scheduler implementation for ILP32 platforms such as watchOS on arm64_32

### Motivation:

Fix a bug in the Scheduler where converting the UInt64 nanoseconds value from a `Duration` value to the `DispatchTimeInterval`'s `Int` associated value causes an uncaught overflow and crash on devices with 32-bit Int values.

### Modifications:

Add `DispatchTime.init(nowDelayedBy:)`, which computes the deadline and avoids casting to `Int`
Replace adhoc computations of Scheduler deadline with designated `DispatchTime.init(nowDelayedBy:)`

### Result:

Prevents crash on ILP32 platforms when time delay exceeds a value that fits within a 32-bit Int

### DIFF
--- a/Sources/DistributedCluster/Scheduler.swift
+++ b/Sources/DistributedCluster/Scheduler.swift
@@ -66,7 +66,7 @@ extension DispatchWorkItem: Cancelable {
 extension DispatchQueue: Scheduler, @unchecked Sendable {
     func scheduleOnce(delay: Duration, _ f: @escaping () -> Void) -> Cancelable {
         let workItem = DispatchWorkItem(block: f)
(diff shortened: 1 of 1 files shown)


ktoso:

ktoso:
Unrelated test hang on 5.8 only: https://github.com/apple/swift-distributed-actors/issues/1138

ktoso:
Thanks, though I'm very surprised about using the cluster on a watch -- that's not a great ideas since it needs to perform heartbeats and active connections which would drain power quickly etc.

Fix looks okey though, thank you!

christopherweems:
Ahh my novice is showing -- thanks for taking a look. I'll look towards moving to another actor system, or rolling one, but it the mean time its really nice to be able to connect to the cluster system while I figure this stuff out. Thanks again!

PR #1139, opened 2023-09-21
Another mini sample to show starting two processes and discovery

ktoso:
To see this in action do:

```
swift run --package-path Samples/ SamplePair first
```
and in another process

```
swift run --package-path Samples/ SamplePair second
```

The actors discover and greet eachother then:

```
===-----------------------------------------------------===
|        Sample App Showing Two Actors Talking            |
|                                                         |
| USAGE: swift run SamplePair first|second                |
===-----------------------------------------------------===
[first-actor on sact://system@127.0.0.1:1111] Received message: Hello
from Person(/user/Person-b["$receptionID": "*", "$path":
/user/Person-b]) on sact://system@127.0.0.1:2222!
[first-actor on sact://system@127.0.0.1:1111] Found:
Person(sact://system@127.0.0.1:2222/user/Person-b)

```

### Motivation:

An even more simple 2 process example to showcase discovery.

### DIFF
--- a/Samples/Sources/SamplePair/boot.swift
+++ b/Samples/Sources/SamplePair/boot.swift
@@ -1,0 +1,59 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Distributed Actors open source project
+//
+// Copyright (c) 2018-2022 Apple Inc. and the Swift Distributed Actors project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.md for the list of Swift Distributed Actors project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Distributed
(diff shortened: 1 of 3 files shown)


Samples/Sources/SamplePair/boot.swift
```diff
@@ -0,0 +1,59 @@
...
+            try! await system.cluster.joined(within: .seconds(10))
+            let p = await Person(name: "first-actor", actorSystem: system)
+
+        default:
+            print("Please select 'first' or 'second'")
+            return
+        }
+
+        try? await Task.sleep(for: .seconds(60))
+        _ = p
```
lyzkov, line 57: `_ = p` changes nothing and means nothing, breaking [Occam's Razor](https://en.wikipedia.org/wiki/Occam%27s_razor) principle. The `_ = p` definition exists only to suppress compiler warnings. Are there any other ways of retaining `Person` instance without declaring an unused constant?

PR #1142, opened 2023-11-06
Lowest address ordering fix

### LINKED ISSUE #1141: Leadership will choose lowest node by `systemName` order and not by address
Current implementation of Comparable protocol from ClusterEndpoint is using description string, and thus creates a problem—it contains `systemName` in it and most important it comes before host and port. So when `selectByLowestAddress` kicks in `Leadership` it will sort nodes and choose lowest not by address, but by alphabetical order (or guess unicode order) of `systemName` node's variables.

As `systemName` is actually internal and not so important quick solution could be just to remove it from node comparison.

akbashev:
### Motivation:

Currently there is a bug in lowest address ordering where it will sort nodes and choose lowest not by address, but by alphabetical order (or guess unicode order) of systemName node's variables.

### Modifications:

Changed implementation of Comparable for ClusterEndpoint and tweaked tests a bit.

### Result:

- Resolves #1141

### DIFF
--- a/Sources/DistributedCluster/ClusterEndpoint.swift
+++ b/Sources/DistributedCluster/ClusterEndpoint.swift
@@ -73,7 +73,7 @@ extension Cluster.Endpoint: Comparable {
     // Silly but good enough comparison for deciding "who is lower node"
     // as we only use those for "tie-breakers" any ordering is fine to be honest here.
     public static func < (lhs: Cluster.Endpoint, rhs: Cluster.Endpoint) -> Bool {
-        "\(lhs)" < "\(rhs)"
+        "\(lhs.protocol)\(lhs.host)\(lhs.port)" < "\(rhs.protocol)\(rhs.host)\(rhs.port)"
     }
 
     public func hash(into hasher: inout Hasher) {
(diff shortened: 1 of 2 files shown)


akbashev:
Event though it's working now properly, I've implemented it naively by string check "\(protocol)\(host)\(port)" and not quite sure about that. 🤔 Maybe there is a better comparison logic? What does mean to check by lowest name of protocol? 🤔

ktoso:
A system will never really have different protocols, so this is okey I think

Sources/DistributedCluster/ClusterEndpoint.swift
```diff
@@ -73,7 +73,7 @@ extension Cluster.Endpoint: Comparable {
     // Silly but good enough comparison for deciding "who is lower node"
     // as we only use those for "tie-breakers" any ordering is fine to be honest here.
     public static func < (lhs: Cluster.Endpoint, rhs: Cluster.Endpoint) -> Bool {
-        "\(lhs)" < "\(rhs)"
+        "\(lhs.protocol)\(lhs.host)\(lhs.port)" < "\(rhs.protocol)\(rhs.host)\(rhs.port)"
```
ktoso, line 76: This is good enough -- just skipping the name 👍

ktoso (approved):
LGTM thanks -- as long as there's a **predictable** order we're good -- so this will do the job. The name should not account for it, good call!

ktoso:

PR #1150, opened 2024-04-16
Added hook to lifecycle events

akbashev:
Ability to listen to actors lifecycles in a plugin.

### Motivation:

Plugins give an option to extend system with some logic. Problem is sometimes you need to hook into actor's creation/deletion, which currently is impossible. As an example—for actor recovery we need to restore state when it's created and ready for cluster system, which can be done by some potential `EventSourcingJournal`.

By adding simple hook protocol we can enable this.

### Modifications:

- `PluginActorLifecycleHook` protocol
- additional `actorLifecycleHooks` in plugins settings
- inside ClusterSystem's `actorReady` and `resignID` are called for each _hookable_ plugin added to the system.

### DIFF
--- a/Sources/DistributedCluster/Plugins/ClusterSystem+Plugins.swift
+++ b/Sources/DistributedCluster/Plugins/ClusterSystem+Plugins.swift
@@ -114,3 +114,9 @@ internal struct AnyPluginKey: Hashable, CustomStringConvertible {
         }
     }
 }
+
+/// Way to hook into ClusterSystem's actor lifecycle events, specifically to `ready` and `resign`.
+public protocol PluginActorLifecycleHook {
+    func actorReady<Act: DistributedActor>(_ actor: Act) where Act: DistributedActor, Act.ID == ClusterSystem.ActorID
+    func resignID(_ id: ClusterSystem.ActorID)
+}
--- a/Sources/DistributedCluster/Plugins/ClusterSystemSettings+Plugins.swift
+++ b/Sources/DistributedCluster/Plugins/ClusterSystemSettings+Plugins.swift
@@ -21,6 +21,7 @@ public struct PluginsSettings {
     }
 
     internal var plugins: [BoxedPlugin] = []
+    internal var actorLifecycleHooks: [PluginActorLifecycleHook] = []
 
     public init() {}
 
@@ -72,7 +73,9 @@ extension PluginsSettings {
             !self.isInstalled(plugin: plugin),
             "Attempted to add plugin \(plugin.key) but key already used! Plugin [\(plugin)], installed plugins: \(self.plugins)."
         )
-
+        if let plugin = plugin as? PluginActorLifecycleHook {
+            self.actorLifecycleHooks.append(plugin)
+        }
         return self.plugins.append(BoxedPlugin(plugin))
     }
 
--- a/Tests/DistributedClusterTests/Plugins/ClusterSingleton/ClusterSingletonPluginTests.swift
+++ b/Tests/DistributedClusterTests/Plugins/ClusterSingleton/ClusterSingletonPluginTests.swift
@@ -59,6 +59,48 @@ final class ClusterSingletonPluginTests: SingleClusterSystemXCTestCase {
         // if this were true we would have crashed by a duplicate name already, but let's make sure:
         singletonID.shouldNotEqual(greeterID)
     }
+    
+    func test_plugin_hooks() async throws {
+        let actorId = "actorHookId"
+        let hookFulfillment = self.expectation(description: "actor-hook")
+        let plugin = TestClusterHookPlugin { actor in
+            /// There are multiple internal actors fired, we only checking for `ActorWithId`
+            guard let actor = actor as? ActorWithId else { return }
+            let id = try? await actor.getId()
+            XCTAssertEqual(id, actorId, "Expected \(actorId) as an id")
+            hookFulfillment.fulfill()
+        }
+        let testNode = await setUpNode("test-hook") { settings in
+            settings.enabled = false
+            settings += plugin
+        }
+
+        let _ = ActorWithId(actorSystem: testNode, id: actorId)
+        await fulfillment(of: [hookFulfillment])
+    }
+    
+    actor TestClusterHookPlugin: _Plugin, PluginActorLifecycleHook {
+        nonisolated var key: Key { "$testClusterHook" }
+        
+        let onActorReady: (any DistributedActor) async throws -> ()
+        
+        init(
+            onActorReady: @escaping (any DistributedActor) async throws -> Void
+        ) {
+            self.onActorReady = onActorReady
+        }
+        
+        nonisolated func actorReady<Act>(_ actor: Act) where Act: DistributedActor, Act.ID == DistributedCluster.ClusterSystem.ActorID {
+            Task { try await self.onActorReady(actor) }
+        }
+        
+        nonisolated func resignID(_ id: DistributedCluster.ClusterSystem.ActorID) {
+            
+        }
+        
+        func start(_ system: ClusterSystem) async throws {}
+        func stop(_ system: ClusterSystem) async {}
+    }
 
     distributed actor SingletonWhichCreatesDistributedActorDuringInit: ClusterSingleton {
         typealias ActorSystem = ClusterSystem
@@ -87,6 +129,22 @@ final class ClusterSingletonPluginTests: SingleClusterSystemXCTestCase {
 
         distributed func greet() {
             print("Hello!")
+        }
+    }
+    
+    distributed actor ActorWithId {
+        let customId: String
+        
+        init(
+            actorSystem: ActorSystem,
+            id: String
+        ) {
+            self.actorSystem = actorSystem
+            self.customId = id
+        }
+        
+        distributed func getId() -> String { 
+            self.customId
         }
     }
 }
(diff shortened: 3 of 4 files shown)


Sources/DistributedCluster/Plugins/ClusterSystem+Plugins.swift
```diff
@@ -114,3 +114,9 @@ internal struct AnyPluginKey: Hashable, CustomStringConvertible {
         }
     }
 }
+
+/// Way to hook into ClusterSystem's actor lifecycle events, specifically to `ready` and `resign`.
+public protocol PluginActorLifecycleHook {
+    func actorReady<Act: DistributedActor>(_ actor: Act) where Act: DistributedActor, Act.ID == ClusterSystem.ActorID
+    func resignID(_ id: ClusterSystem.ActorID)
+}
```
akbashev, line 122: Explicitly made non as it's called synchronously in system.
This means though that if journal is an actor—those functions will be nonisolated. Thinking now if it's ok or not 🤔

akbashev:
Still not sure about `PluginActorLifecycleHook ` name 😅

Sources/DistributedCluster/Plugins/ClusterSystem+Plugins.swift
ktoso: yeah those funcs being synchronous is good -- the plugin may have to use external synchronization, that's ok I guess -- we don't have a choice since the hooks are synchronous.

Naming nitpick, let's call them `onActorReady` and `onResignID`

Sources/DistributedCluster/Plugins/ClusterSystemSettings+Plugins.swift
```diff
@@ -21,6 +21,7 @@ public struct PluginsSettings {
     }
 
     internal var plugins: [BoxedPlugin] = []
+    internal var actorLifecycleHooks: [PluginActorLifecycleHook] = []
```
ktoso, line 24: the front of the name is good; but maybe `PluginActorLifecyclePlugin`?

Tests/DistributedClusterTests/Plugins/ClusterSingleton/ClusterSingletonPluginTests.swift
```diff
@@ -59,6 +59,48 @@ final class ClusterSingletonPluginTests: SingleClusterSystemXCTestCase {
...
+        let testNode = await setUpNode("test-hook") { settings in
+            settings.enabled = false
+            settings += plugin
+        }
+
+        let _ = ActorWithId(actorSystem: testNode, id: actorId)
+        await fulfillment(of: [hookFulfillment])
+    }
+    
+    actor TestClusterHookPlugin: _Plugin, PluginActorLifecycleHook {
```
ktoso, line 82: Oh plugin was hidden still huh. About time to make it public -- maybe as we cleanup the how we use them with your work here!

Sources/DistributedCluster/Plugins/ClusterSystem+Plugins.swift
```diff
@@ -114,3 +114,9 @@ internal struct AnyPluginKey: Hashable, CustomStringConvertible {
         }
     }
 }
+
+/// Way to hook into ClusterSystem's actor lifecycle events, specifically to `ready` and `resign`.
+public protocol PluginActorLifecycleHook {
```
ktoso, line 119: Should this extend `_Plugin`?

Tests/DistributedClusterTests/Plugins/ClusterSingleton/ClusterSingletonPluginTests.swift
```diff
@@ -89,4 +131,20 @@ final class ClusterSingletonPluginTests: SingleClusterSystemXCTestCase {
             print("Hello!")
         }
     }
+    
+    distributed actor ActorWithId {
```
ktoso, line 135:
```suggestion
    distributed actor ActorWithID {
```

Tests/DistributedClusterTests/Plugins/ClusterSingleton/ClusterSingletonPluginTests.swift
```diff
@@ -89,4 +131,20 @@ final class ClusterSingletonPluginTests: SingleClusterSystemXCTestCase {
...
+        
+        init(
+            actorSystem: ActorSystem,
+            id: String
+        ) {
+            self.actorSystem = actorSystem
+            self.customId = id
+        }
+        
+        distributed func getId() -> String {
```
ktoso, line 146:
```suggestion
        distributed func getID() -> String { 
```

big `ID` everywhere please :)

Tests/DistributedClusterTests/Plugins/ClusterSingleton/ClusterSingletonPluginTests.swift
```diff
@@ -59,6 +59,48 @@ final class ClusterSingletonPluginTests: SingleClusterSystemXCTestCase {
...
+        let onActorReady: (any DistributedActor) async throws -> ()
+        
+        init(
+            onActorReady: @escaping (any DistributedActor) async throws -> Void
+        ) {
+            self.onActorReady = onActorReady
+        }
+        
+        nonisolated func actorReady<Act>(_ actor: Act) where Act: DistributedActor, Act.ID == DistributedCluster.ClusterSystem.ActorID {
+            Task { try await self.onActorReady(actor) }
```
ktoso, line 94: probably make use of a lock inside here and make it a class to avoid the racy nature of Task {}

Sources/DistributedCluster/Plugins/ClusterSystem+Plugins.swift
```diff
@@ -114,3 +114,9 @@ internal struct AnyPluginKey: Hashable, CustomStringConvertible {
         }
     }
 }
+
+/// Way to hook into ClusterSystem's actor lifecycle events, specifically to `ready` and `resign`.
```
ktoso, line 118:
```suggestion
/// Kind of `ClusterSystem` plugin which will be invoked during an actor's `actorReady`
/// and `resignID` lifecycle hooks.
///
/// The ready hook is allowed to modify the ID, e.g. by adding additional metadata to it.
/// The plugin should carefully manage retaining actors and document if it does have strong references to them,
/// and how end-users should go about releasing them.
```

ktoso (approved):
Very good! 

Minor nitpicks.

Documentation we can add once we document _Plugin itself I suppose :)

Tests/DistributedClusterTests/Plugins/ClusterSingleton/ClusterSingletonPluginTests.swift
akbashev: 🫡
Will remove `_` from Plugin

Sources/DistributedCluster/Plugins/ClusterSystemSettings+Plugins.swift
akbashev: Still not sure, maybe more like `ActorLifecycleAwarePlugin`? 🤔

Sources/DistributedCluster/Plugins/ClusterSystem+Plugins.swift
akbashev: Ah, so your idea is that it's still a plugin. Hm 🤔 I was more going into direction if you want to extend your current plugin, you can just conform it to additional protocol, like `class CustomPlugin: Plugin, ActorLifecyleHookable` or something. 
Though also making it directly a plugin also makes sense, you basically distinguish between normal and hookable plugins 🤔

Tests/DistributedClusterTests/Plugins/ClusterSingleton/ClusterSingletonPluginTests.swift
akbashev: This is why I was worried about sync functions in protocol. Working with actors you'll mostly land in `async` anyway or your plugin could be an actor. And then you'll have `onisolated func actorReady` with `Task` or some later stage.
Which is also ok though if you handle it inside your plugin I guess 🤔

Sources/DistributedCluster/Plugins/ClusterSystemSettings+Plugins.swift
akbashev: or just `LifecyclePlugin`... 🤔

akbashev: I see usage of `ActorLifecycle` term in the system, so will just call `ActorLifecyclePlugin`.

ktoso:

ktoso:

ktoso:

akbashev:

Tests/DistributedClusterTests/Plugins/ClusterSingleton/ClusterSingletonPluginTests.swift
ktoso: I know this is a very old discussion, but to clarify: the actor ready must be synchronous because we support synchronous init() in actors.

akbashev:
@ktoso thx for merging main, fixed issue with soundness, now should be fine

ktoso:
Xcode builds consistently timeout: 

>   failed downloading 'https://github.com/apple/swift-protobuf/releases/download/protoc-artifactbundle-v31.1/protoc-31.1.artifactbundle.zip' which is required by binary target 'protoc': downloadError("The request timed out.")

PR #1155, opened 2024-06-14
ClusterD PoC which serves as seed node

ktoso:
Quick PoC how a `clusterd` daemon could look like.

Usage in practice would be to just ensure we start the daemon OR if ANY of the nodes is using daemon discovery make sure one of them spawns it or something like that.

This way nodes can join without hardcoding discovery of peer nodes when playing around locally.
This is inspired by https://forums.swift.org/t/beginner-advice-for-distributed-distributedcluster-modules/72391/11 and erlang's shortnames + https://www.erlang.org/docs/25/man/epmd

---

Creating a few nodes in a system would literarily be:

```bash 
> clusterd & # or as a system-service - which is a binary we'd prepare and basically just does
``` 
this basically just does:
```
//     await ClusterSystem.startClusterDaemon()
```

This is the same as epmd, it has to be running somewhere.

> EPMD is an external program, [implemented in C](https://github.com/erlang/otp/blob/maint/erts/epmd/src/epmd.c). Whilst net_kernel:start/1 takes care of [creating the net_sup supervisor](https://github.com/erlang/otp/blob/maint/lib/kernel/src/erl_distribution.erl#L37), it does not actually trigger the EPMD daemon, which has to be started explicitely. I had a look at how EPMD is started when the -sname option is specified in the erl command and - surprise, surprise - I discovered that [the epmd program is started via a system() C call](https://github.com/erlang/otp/blob/maint/erts/etc/common/erlexec.c#L1167-L1209).

And applications just do:

```
    let system = await ClusterSystem(name) { settings in
      settings.discovery = .clusterd
    }
```

and you're done:

```bash
> myapp --name first
> myapp --name second
```

they'll join eachother.

Follow up here: 

- [ ] `ClusterD` should be a "HIDDEN" member in the membership
- [ ] Maybe we should call it `swift-clusterd` (the cluster daemon for the swift distributed cluster)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -109,9 +109,16 @@ var targets: [PackageDescription.Target] = [
             // Depend on tests to run:
             "DistributedActorsMultiNodeTests",
 
-            // Dependencies:
             "MultiNodeTestKit",
             .product(name: "ArgumentParser", package: "swift-argument-parser"),
+        ]
+    ),
+
+    .executableTarget(
+        name: "Clusterd",
+        dependencies: [
+            "DistributedCluster",
+           .product(name: "ArgumentParser", package: "swift-argument-parser"),
         ]
     ),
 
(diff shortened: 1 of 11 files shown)


akbashev (commented):
Some formatting issues (guess better run swift format) and nitpicky comments, otherwise really like the idea.

As a follow ups would be nice to have it in docs and maybe some small demo.

PR #1158, opened 2024-07-03
Addressing warnings

### LINKED ISSUE #1125: Must cleanup amount of warnings in 5.9 build
We have a lot of self induced "deprecation warnings" because we use the runtime that people should not be touching.

- We should probably remove a bunch of those deprecations because they are not types the public API exposes and we just have the noise to suffer through it.
- We should consistently remove those deprecated "actor ref" APIs
- We should address new warnings that appear in the 5.9 build

Garfeild:
### Motivation:

Currently project is swamped with warnings, especially due to `Never` getting conformance to `Codable` in newer macOS versions.

### Modifications:

This pull request contains following modifications:
* Bump of the supported macOS version to `.v14`
* set of non-functional changes to address majority of warnings 

### Result:

- Partially addresses #1125

| Before changes | After changes |
| --- | --- |
| Screenshot 2024-07-03 at 06 18 56 | Screenshot 2024-07-03 at 06 19 42


Garfeild:

Sources/DistributedCluster/ClusterSystem.swift
```diff
@@ -232,11 +232,6 @@ public class ClusterSystem: DistributedActorSystem, @unchecked Sendable {
         var settings = settings
         self.name = settings.endpoint.systemName
 
-        // rely on swift-backtrace for pretty backtraces on crashes
-        if settings.installSwiftBacktrace {
-            Backtrace.install()
-        }
```
ktoso, line 238: Good to remove indeed, thanks

ktoso (approved):
Looks good, thank you!

ktoso:

ktoso:

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.7
+// swift-tools-version:5.10
```
ktoso, line 1: We're technically supporting 5.7+... To be honest I think it might be good to bump the requirement here since we're able to do so actually given current users.

We might need to think about requiring 6.0 soon though, because of new language features we'd like to use, but we'll need a stable release.

ktoso:

Package.swift
Garfeild: Yeah, we couldn't bump minimal macOS version without bumping swift-tools as well, it should be at least `5.9` 😔

ktoso: Sorry for the long delay here, I'm planning to get this merged now very soon. Travel and sickness got in the way, sorry for the wait

ktoso:

ktoso:
Trying to land this; we need soundness to not fail though

Garfeild:
Thanks for getting back! I will check it. 

And no worries about delays!

ktoso:
`18:16:50 error: 'code': package 'code' is using Swift tools version 5.10.0 but the installed version is 5.7.3` I see... I'll see about updating our CI configurations, sorry for the wait.

ktoso:

ktoso:

ktoso:

ktoso:
Okey we're good here and we changed CI around quite a bit -- we'll just do 5.10+ from now on.

ktoso:
Thank you for your work here @Garfeild !

PR #1159, opened 2024-07-23
poc of distributed actor gossiper; the algorithm can be improved independent of user code

ktoso:
Second commit here shows off how we can implement a simple gossip mechanism.

This is just a quick PoC not a complete implementation and implemented as a sample rather than the core of the library for now, but we should move that over.

### DIFF
--- a/Samples/Sources/SimpleGossip/Gossiper.swift
+++ b/Samples/Sources/SimpleGossip/Gossiper.swift
@@ -1,0 +1,112 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Distributed Actors open source project
+//
+// Copyright (c) 2018-2022 Apple Inc. and the Swift Distributed Actors project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.md for the list of Swift Distributed Actors project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import DistributedCluster
+import Logging
+
+protocol GossipAcknowledgement<Payload>: Sendable, Codable{
+  associatedtype Payload where Payload: Sendable & Codable
+
+  var sequenceNumber: UInt { get }
+
+  // Optional support for simple push-pull gossip;
+  // when we receive an "old" version, we can immediately reply with a newer one
+  var supersededBy: Payload? { get }
+}
+
+protocol Gossiper<Gossip, Acknowledgement>: DistributedActor where ActorSystem == ClusterSystem {
+  associatedtype Gossip: Codable
+  associatedtype Acknowledgement: GossipAcknowledgement<Gossip>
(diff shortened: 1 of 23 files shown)


Samples/Sources/SimpleGossip/Gossiper.swift
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  func gossipRound() async throws {
+    guard let target = self.gossipPeers.shuffled().first else {
+      log.info("No peer to gossip with...")
+      return
+    }
+
+    guard target.id != self.id else {
+      return try await gossipRound() // try again
```
akbashev, line 102: Also while testing and while cluster is forming this function goes in loop heavily 🤔 maybe adding something like:
```suggestion
     if self.gossipPeers.count == 1 {
         try await Task.sleep(for: .seconds(0.1))
     }
      return try await gossipRound() // try again
```

Samples/Sources/SimpleGossip/GreetingGossip.swift
```diff
@@ -0,0 +1,79 @@
...
+    self.activeGreeting = gossip
+
+    return .init(sequenceNumber: gossip.sequenceNumber)
+  }
+
+//    distributed func acknowledgement(_ acknowledgement: Acknowledgement,
+//                                from peer: ID,
+//                                confirming gossip: Gossip) {
+//
+//    }
```
akbashev, line 78: Is it needed? Not like nitpicking, but rather thinking about future refactoring—I've noticed there are some commented code here and there (not much though) and sometimes not sure if I should just remove it or not.

Sources/DistributedCluster/InvocationBehavior.swift
```diff
@@ -34,7 +34,7 @@ public struct InvocationMessage: Sendable, Codable, CustomStringConvertible {
 
 // FIXME(distributed): remove [#957](https://github.com/apple/swift-distributed-actors/issues/957)
 enum InvocationBehavior {
-    static func behavior(instance weakInstance: Weak<some DistributedActor>) -> _Behavior<InvocationMessage> {
+    static func behavior(instance weakInstance: WeakLocalRef<some DistributedActor>) -> _Behavior<InvocationMessage> {
```
akbashev, line 37: Wonder if we can just remove WeakLocalRef here and just check for weak in closures bellow.

Sources/DistributedCluster/Pattern/WorkerPool.swift
```diff
@@ -200,6 +246,30 @@ extension WorkerPool {
             .init(underlying: .static(workers))
         }
     }
+
+    public struct Strategy {
+        enum _Strategy {
+            case random
+            case simpleRoundRobin
+        }
```
akbashev, line 254: Nice!

akbashev (commented):
Some small comments from my side.

imho should be merged even in the form of PoC, benefits are big and we can figure out later on how to improve.

PR #1161, opened 2024-09-19
Made ClusterSingletonSettings sendable

akbashev:
Using ClusterSingleton's host function with latest strict concurrency will result in an error:
```
Sending 'self'-isolated value of type 'ClusterSingletonSettings' with later accesses to actor-isolated
context risks causing data races
```

As ClusterSingletonPlugin is an actor and settings cross isolation in host function—it's better, and think safe, to mark it as Sendable.


ktoso (approved):

ktoso:
Thank you!

PR #1164, opened 2024-10-15
Fixed WorkerPool to handle remote actors properly

akbashev:
Currently WorkerPool:

- Treats remote actors as weak and cleans them directly
- Doesn't work on Swift 6 mode as `WorkItem` and `WorkResult` are not marked as `Sendable`

This PR fixes this, basically cherry picked changes from this PR https://github.com/apple/swift-distributed-actors/pull/1159 and updated tests


Sources/DistributedCluster/Pattern/WorkerPool.swift
```diff
@@ -53,33 +54,37 @@ public distributed actor WorkerPool<Worker: DistributedWorker>: DistributedWorke
...
+        self.actorSystem = system
+        
+        switch settings.selector {
         case .dynamic(let key):
             self.newWorkersSubscribeTask = Task {
                 for await worker in await self.actorSystem.receptionist.listing(of: key) {
-                    self.actorSystem.log.log(level: self.logLevel, "Got listing member for \(key): \(worker)")
-                    self.workers[worker.id] = Weak(worker)
+                    self.workers.append(WeakLocalRef(worker))
+                    self.workers.sort(by: { $0.id < $1.id })
```
akbashev, line 82: Currently for test to pass I'm sorting workers every time, atm not sure of OrderSet brings any value 🤔
need to double check

Sources/DistributedCluster/Pattern/WorkerPool.swift
```diff
@@ -172,21 +216,12 @@ extension WorkerPool {
...
     /// using this key to the pool. On the other hand, static configurations could restrict the set
     /// of members to be statically provided etc.
-    public struct Selector {
-        enum _Selector {
-            // TODO: let awaitAtLeast: Int // before starting to direct traffic
-            case dynamic(DistributedReception.Key<Worker>)
-            case `static`([Worker])
-        }
-
-        let underlying: _Selector
```
akbashev, line 182: @ktoso I wonder why `underlying` is needed and not to use enum directly? I've updated it to be enum, but maybe there are some concerns...

akbashev:
Hm, would be nice to add test to check remote actors also.

Sources/DistributedCluster/Pattern/WorkerPool.swift
ktoso: Don't do public enums -- we can never add a case to them basically because of source compatibility. So this has to remain as a struct that hides the enum inside it

Sources/DistributedCluster/Pattern/WorkerPool.swift
```diff
@@ -196,9 +229,21 @@ extension WorkerPool {
...
+
+        /// Round-robin strategy which attempts to go "around" known workers one-by-one
+        /// giving them equal amounts of work. This strategy is NOT strict, and when new
+        /// workers arrive at the pool it may result in submitting work to previously notified
+        /// workers as the round-robin strategy "resets".
+        ///
+        /// We could consider implementing a strict round robin strategy which remains strict even
+        /// as new workers arrive in the pool.
+        case simpleRoundRobin
     }
```
ktoso, line 247: Again we should make a struct with an underlying enum inside it, and make static functions to create instances. Otherwise by making the enum public we'll never be able to add new cases without a major release.

ktoso (commented):
Looks good, just needs to be careful with the enums and maybe add one more test?

Sources/DistributedCluster/Pattern/WorkerPool.swift
akbashev: ah, good point, haven't thought about this.

ktoso:

ktoso:

ktoso:

ktoso (approved):

PR #1171, opened 2024-10-25
PoC: DistributedSequence that is DistributedActorSystem agnostic!

ktoso:
This is a simple, yet proving the point, implementation of a `DistributedSequence<Element, ActorSystem>`!

Thanks to advances in most recent Swift we're able to express such generic over actor system protocols and provide general utility libraries which work with any distributed actor system.

Implementation details are simple in this one, we'd need to be a bit smarter with potentially batching (high-low watermark fetching) as well as timeouts and failure handling of the subscriber (when the subscriber node dies). But this in general proves the mechanisms that one would use to implement such type.

Since it is `ClusterSystem` agnostic, we could make this a general type in `Distributed` module!

### DIFF
--- a/Sources/DistributedCluster/DistributedSequence.swift
+++ b/Sources/DistributedCluster/DistributedSequence.swift
@@ -1,0 +1,151 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Distributed Actors open source project
+//
+// Copyright (c) 2018-2022 Apple Inc. and the Swift Distributed Actors project authors
+// Licensed under Apache License v2.0
+//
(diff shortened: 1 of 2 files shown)


ktoso:
TODO, maybe make it single subscriber

TODO, if it's possible to do a proxy with child tasks, like these in some other libs

```
func stream(body: (some AsyncWriter<Int>, some AsyncSequence<Int>) async -> Void) async {
  await withTaskGroup { group in
    let writer = ...
    let sequence = ...
    group.addTask {
      // Pull data from writer and push data to sequence
    }
   await body(writer, sequence)
  }
}
```

Issue #1172, opened 2024-11-01
Require Swift 6


ktoso:
Due to some memory leak issues in remote calls that only got fixed in Swift 6 we are considering to require the use of Swift 6 in this library. 

Usage is relatively low, and all adopters I'm aware of are okey to update their Swift dependency (or already have).

If someone with production use has concerns with this please let us know.

PR #1173, opened 2024-11-02
Require Swift 6

akbashev:
This PR resolves this issue https://github.com/apple/swift-distributed-actors/issues/1172

Before merging we need to do two things though:

- Latest Xcode 16.1 still having older Swift version and memory leak not yet fixed, for better cross-platform experience we need to wait until next version is released.
- [Tests are failing with timeout](https://ci.swiftserver.group/job/swift-distributed-actors-6.0-prb/)—really strange issue. Needs to be fixed as Swift 6 will become required.


Package.swift
```diff
@@ -227,10 +237,10 @@ platforms = nil
...
 platforms = [
     // we require the 'distributed actor' language and runtime feature:
-    .iOS(.v16),
-    .macOS(.v14),
-    .tvOS(.v16),
-    .watchOS(.v9),
+    .iOS(.v18),
+    .macOS(.v15),
+    .tvOS(.v18),
+    .watchOS(.v11),
```
akbashev, line 243: While this is not required, it will give some goodies like Synchronization

Package.swift
```diff
@@ -44,7 +44,8 @@ var targets: [PackageDescription.Target] = [
             .product(name: "ServiceDiscovery", package: "swift-service-discovery"),
             .product(name: "Backtrace", package: "swift-backtrace"),
             .product(name: "AsyncAlgorithms", package: "swift-async-algorithms"),
-        ]
+        ],
+        swiftSettings: [.swiftLanguageMode(.v5)]
```
akbashev, line 48: Decided to put it per target for future migration to strict concurrency

akbashev:
Just realised we need to update Readme (and probably docs)

Package.swift
ktoso: Agreed, we’ll go one by one here

ktoso:

ktoso:

akbashev:
@ktoso ah, sorry, mistakenly asked to review wrong PR. This is too early.

ktoso:
No problem :)

Package.swift
ktoso: that's ok

ktoso (approved):

ktoso:
I think we could also disable the cxx interop jobs tbh, they're not giving much real signal for us

PR #1177, opened 2024-11-19
Enable shell check and fix linting

rnro:
Enable shell check and fix linting.

### Motivation:

More soundness checks.

### Modifications:

Enable shell check and fix linting.

### Result:

More soundness checks.


FranzBusch:
Merging over this since the tests on 6.0 have issues

PR #1189, opened 2025-03-14
Run all non-soundness CI jobs on Main

czechboy0:
### Motivation:

Main jobs run nightly, and allow us to catch when things regress. Right now we're catching issues on PRs only.

### Modifications:

Add the missing jobs to main.

### Result:

Better CI coverage on main, easier bisecting.


czechboy0:
Again, merging over red CI to fix up Main CI.

Issue #1179, opened 2024-11-20
Tests crash and unit tests hang on Swift 6.0


rnro:
See e.g.
https://github.com/apple/swift-distributed-actors/actions/runs/11934835128/job/33264822358?pr=1178
https://github.com/apple/swift-distributed-actors/actions/runs/11934835128/job/33264815917?pr=1178

```
Test Case 'SerializationTests.test_serialize_Int_withData' started at 2024-11-20 14:01:22.517
Swift/UnsafeRawPointer.swift:440: Fatal error: load from misaligned raw pointer

*** Signal 4: Backtracing from 0x7fc4a8c02bb4... done ***

*** Program crashed: Illegal instruction at 0x00007fc4a8c02bb4 ***

Thread 0 "swift-distribut" crashed:

0              0x00007fc4a8c02bb4 closure #1 in closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) + 308 in libswiftCore.so
1 [ra]         0x00007fc4a8c02962 closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) + 257 in libswiftCore.so
2 [ra]         0x00007fc4a8c0281d closure #1 in _assertionFailure(_:_:file:line:flags:) + 380 in libswiftCore.so
3 [ra]         0x00007fc4a8d85d16 _assertionFailure(_:_:file:line:flags:) + 229 in libswiftCore.so
4 [ra]         0x00007fc4a8d85f6e _fatalErrorMessage(_:_:file:line:flags:) + 45 in libswiftCore.so
5 [ra]         0x00007fc4a8e07bd0 UnsafeRawBufferPointer.load<A>(fromByteOffset:as:) + 191 in libswiftCore.so
6 [ra]         0x00005626739d427d closure #1 in IntegerSerializer.deserialize(from:) + 92 in swift-distributed-actorsPackageTests.xctest at /__w/swift-distributed-actors/swift-distributed-actors/Sources/DistributedCluster/Serialization/Serialization+PrimitiveSerializers.swift:76:64
7 [ra] [thunk] 0x00005626739d4a28 partial apply for closure #1 in IntegerSerializer.deserialize(from:) + 23 in swift-distributed-actorsPackageTests.xctest at //<compiler-generated>
8 [ra]         0x00007fc4a95c43bf partial apply for closure #1 in Data.InlineData.withUnsafeBytes<A>(_:) + 30 in libFoundationEssentials.so
[...]

PR #1190, opened 2025-03-17
Fixed a crash on Swift 6.0+ where Integer deserialization was assuming aligned memory

czechboy0:
### Motivation:

The integer deserializer was assuming that the input raw memory was correctly aligned for the integer type to be deserialized. In the test that was true on some platforms, but not others.

This caused a crash in the unit test.

### Modifications:

Use `loadUnaligned` instead of `load` when loading integers.

### Result:

Fixes https://github.com/apple/swift-distributed-actors/issues/1179.


simonjbeaumont (approved):

czechboy0:
The various failing CI jobs are all caused by different issues. Merging over red CI because this PR fixes one of them, will continue in subsequent PRs.

PR #1202, opened 2025-04-10
Switch integration tests to newer Swift test matrix

rnro:
### Motivation:

To remove all uses of the older workflow so it can be removed to simplify the workflows.

### Modifications:

* Switch integration tests an the dining philosophers example to use the newer Swift test matrix workflow, `swift_test_matrix.yml`

### Result:

This should make no difference to the running of integration tests.


czechboy0 (approved):
Thanks, @rnro!

PR #1205, opened 2025-04-30
Fix the Samples project so that it compiles and runs.

zaneenders:
### Motivation:

Currently, the sample project doesn't compile because `swift-cluster-membership` requires macos 14.0.

```console
error: the executable 'SampleDiningPhilosophers' requires macos 13.0, but depends on the product 'DistributedCluster' which requires macos 14.0; consider changing the executable 'SampleDiningPhilosophers' to require macos 14.0 or later, or the product 'DistributedCluster' to require macos 13.0 or earlier.
```

### Modifications:

Update the `Package.swift` file in `/Samples` to require macos 14 and bump to Swift 5.9 to support running in distributed mode.

Updated README to include `swift` in the run commands.

### Result:

`Samples` project now compiles and runs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ You can refer to the `Samples/` directory to view a number of more realistic sam
 
 The most "classical" example of distributed actors is the [`SampleDiningPhilosophers`](Samples/Sources/SampleDiningPhilosophers/DistributedDiningPhilosophers.swift).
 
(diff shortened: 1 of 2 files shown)


ktoso:
Thanks!

ktoso (approved):

PR #1208, opened 2025-05-08
Fix hang using multi-node package plugin on Swift 6+

simonjbeaumont:
### Motivation:

The integration tests for Swift 6.0 and 6.1 are hanging. They are passing on Swift 5.10. This is because SwiftPM started taking a lock over the package and cannot be run concurrently on the same package.

This is causing hangs in the integration test because it calls the `multi-node` package plugin which is running the `MultiNodeTestKitRunner` using `swift run MultiNodeTestKitRunner`.

This isn't the supported way for Swift package plugins to use executables from the package, which should instead be looked up from the context passed into the plugin.

### Modifications:

Use the package plugin context to find the executable dependency used by the multi-node plugin.

### Result:

Integration tests no longer hang on Swift 6.0 and 6.1.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -93,7 +93,7 @@ var targets: [PackageDescription.Target] = [
             // permissions: needs full network access
         ),
         dependencies: [
-            "MultiNodeTestKitRunner",
+            "MultiNodeTestKitRunner"
         ]
     ),
     .target(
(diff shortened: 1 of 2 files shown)


simonjbeaumont:
@czechboy0 @rnro this is a false positive from the API breakage checker, right? Can't see how it could be legit.

PR #1209, opened 2025-05-08
Fix compilation on platforms without Foundation.Process

simonjbeaumont:
### Motivation:

There's an open PR to add CI for Apple platforms. This is currently failing since it tries to build for all Apple platforms, not just macOS. This is because there are targets that provide support for testing that rely on `Foundation.process`, which is unavailable on non-macOS Apple platforms.  

### Modifications:

Guard the use of `Foundation.Process` with an `#if canImport`. 

### Result:

Package builds for non-macOS Apple platforms: iOS, tvOS, watchOS, etc.

### DIFF
--- a/Sources/DistributedActorsTestKit/InspectKit.swift
+++ b/Sources/DistributedActorsTestKit/InspectKit.swift
@@ -22,6 +22,7 @@ internal struct InspectKit {
         FileManager.default.currentDirectoryPath
     }
 
+    #if canImport(Foundation.Process)
     private static func runCommand(cmd: String, args: String...) -> (output: [Substring], error: [Substring], exitCode: Int32) {
         var output: [Substring] = []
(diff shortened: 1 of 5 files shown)


rnro (approved):

simonjbeaumont:
Looks like a flakey test struggling to establish the cluster before terminating it, timing out after 5 seconds. 

```
Test Case 'ShootTheOtherNodeClusteredTests.test_shootOtherNodeShouldTerminateIt' started at 2025-05-14 08:35:08.125
<EXPR>:0: error: ShootTheOtherNodeClusteredTests.test_shootOtherNodeShouldTerminateIt : threw error "
        try await self.joinNodes(node: local, with: remote, ensureWithin: .seconds(5), ensureMembers: .up)
```

Runs locally without reproducing 100 times in a row.

PR #1210, opened 2025-05-13
disable swift test on macOS

rnro:
Disable swift test on macOS as the tests often hang https://github.com/apple/swift-distributed-actors/actions/runs/14981553866/job/42086681590


ktoso:
"hang" fails language test

ktoso (approved):

PR #1165, opened 2024-10-15
Finishing clusterd

akbashev:
Quite often people ask "How actors find each other" and clusterd could be a good example to show discoverability.

Again just stealing your code [here](https://github.com/apple/swift-distributed-actors/pull/1155) @ktoso 🙂 Will just try to finish and prettify.

PR #1213, opened 2025-05-14
Disable api breakage check

akbashev:
https://github.com/apple/swift-distributed-actors/pull/1165#issuecomment-2844558188
We can disable API breakage check for now as we're pre-1.0


akbashev:
@simonjbeaumont just seen your [PR](https://github.com/apple/swift-distributed-actors/pull/1211), so probably not so relevant anymore?

simonjbeaumont:
Correct. It's also now passing with an empty PR: #1212. I've raised this issue with the SwiftPM folks.

Issue #1223, opened 2025-09-24
Sending value of non-Sendable type '(inout ClusterSystemSettings) -> Void' risks causing data races


akbashev:
## Environment 
macOS 26, Xcode 26.0.1, Swift 6.2 (6.2.0.19.9)
App: Simple executable with in _swift-tools-version: 6.0_ package without any additional swift settings.

## Description
When creating a cluster system on main actor compiler will fire an error, e.g.:
```swift
@main
struct App {
  static func main() async throws {
    let someActorsNode = await ClusterSystem("some_actors") { // Sending value of non-Sendable type '(inout ClusterSystemSettings) -> Void' risks causing data races
      $0.bindHost = "127.0.0.1"
      $0.bindPort = 2559
    }
    try await someActorsNode.terminated
  }
}
```
Error:
```
Sending value of non-Sendable type '(inout ClusterSystemSettings) -> Void' risks causing data races
```
<img width="1197" height="390" alt="Image" src="https://github.com/user-attachments/assets/54913c36-06e9-43b2-960d-5584a4982ccf" />

## Swift issue?
Interestingly enough when creating a main.swift file and executing at top level (`without @main struct App {`) compiler won't complain, but app will crash when executing configureSettings:
```
warning: TypeSystemSwiftTypeRef::operator(): had to engage SwiftASTContext fallback for type $s18DistributedCluster0B14SystemSettingsVD  thread #1
    frame #0: 0x000000018f8b1c34 libsystem_kernel.dylib`mach_msg2_trap + 8
    frame #1: 0x000000018f8c4028 libsystem_kernel.dylib`mach_msg2_internal + 76
    frame #2: 0x000000018f8ba98c libsystem_kernel.dylib`mach_msg_overwrite + 484
    frame #3: 0x000000018f8b1fb4 libsystem_kernel.dylib`mach_msg + 24
    frame #4: 0x000000018f993c80 CoreFoundation`__CFRunLoopServiceMachPort + 160
    frame #5: 0x000000018f9925d8 CoreFoundation`__CFRunLoopRun + 1188
    frame #6: 0x000000018fa50898 CoreFoundation`_CFRunLoopRunSpecificWithOptions + 532
    frame #7: 0x000000018f9e9a44 CoreFoundation`CFRunLoopRun + 64
    frame #8: 0x0000000287fd7ee8 libswift_Concurrency.dylib`CFMainExecutor.run() + 48
[...]

Issue #1215, opened 2025-05-31
Migrate Cluster system to Swift structured concurrency


akbashev:
The cluster system was written before swift structured concurrency was introduced and actually uses its own concurrency runtime. That works fine, but adds some complexity and introduces problems for further development—one must keep two concepts in mind and provide ways to glue them together. Not to mention keeping dedicated concurrency parts and having to deal with two simultaneously is less scalable.

Natural solution for that is to move library to structured concurrency. There are two possible ways to start this migration:
- Remove old concurrency runtime step by step, class by class, and replace it with new structure concurrency.
- Rewrite system from scratch and use structured concurrency directly.

First way seems safe, but not straightforward—even though old actor references and behavior related files are marked as deprecated, system is built on top of those abstractions. So touching any of them leads to massive chain of changes. 

Second seems massive and breaking, but actually way more straightforward—we already have clear picture of the system, and just need to rewrite/refactor to desired state. This will also give opportunity to refactor related repositories (e.g. swift-cluster-membership) with cleaner APIs. 

After talking with @ktoso he proposed going full rewrite way with following steps:

- [x] Start a new system implementation, just doing the remote call. Static list of nodes, no actor references or behaviors.
- [ ] Make SWIM (swift-cluster-membership) just async and remove protocols from it.
- [ ] Adopt in new cluster membership into the system for failure detection.
- [ ] Use some consensus (e.g., Raft) for the membership.
- [ ] Plugin infrastructure, so we have can have the singleton and ability to extend system (event sourcing, virtual actors).
- [ ] For the network take care about the nodes joining each other—there’s a race there.
- [ ] Modernize the NIO networking layer with Swift concurrency features.
- [ ] (optional) Add tracing for easier debugging and observability.

Note that this is all high level, so each step could be clarified if needed. Also, even though all steps are high level—this is all part of one big refactor, so careful approach should be considered for refactoring and submitting PRs.

akbashev:
> so careful approach should be considered for refactoring and submitting PRs.
@ktoso can we create a separate `refactoring` branch in repo for that? My idea is that I'll create a NewClusterSystem package in it and gradually will move/delete old files from old system and create PRs for each step.

akbashev:
@AustinSnooz yeah, sure 🙂 the only thing we need to think about is how to manage it, can you dm me (email of social networks)?

PR #1224, opened 2025-09-25
Mark configuration setting closure as sending

akbashev:
### Motivation:

Configuring cluster system on main actor (whoever needs this) using configuration closure init leads to compiler error `Sending value of non-Sendable type '(inout ClusterSystemSettings) -> Void' risks causing data races`.

### Modifications:

As library already been updated to Swift 6 and ClusterSystemSettings is not used outside configuration closure — feels like marking closure as **sending** is more suitable than using **Sendable** conformance.

### Result:

- Resolves [#1223](https://github.com/apple/swift-distributed-actors/issues/1223)

### Additional information

There are other same type closures in the system, but as they're not exposed to outside world — will be fixed [during refactoring](https://github.com/apple/swift-distributed-actors/issues/1215).

### DIFF
--- a/Sources/DistributedCluster/Behaviors.swift
+++ b/Sources/DistributedCluster/Behaviors.swift
@@ -131,10 +131,11 @@ extension _Behavior {
     func receiveSignalAsync(
         context: _ActorContext<Message>,
         signal: _Signal,
-        handleSignal: @escaping @Sendable (
-            _ActorContext<Message>,
-            _Signal
-        ) async throws -> _Behavior<Message>
+        handleSignal:
+            @escaping @Sendable (
+                _ActorContext<Message>,
+                _Signal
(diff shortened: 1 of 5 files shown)


ktoso (approved):

Sources/DistributedCluster/Behaviors.swift
```diff
@@ -131,10 +131,11 @@ extension _Behavior {
...
         signal: _Signal,
-        handleSignal: @escaping @Sendable (
-            _ActorContext<Message>,
-            _Signal
-        ) async throws -> _Behavior<Message>
+        handleSignal:
+            @escaping @Sendable (
+                _ActorContext<Message>,
+                _Signal
+            ) async throws -> _Behavior<Message>
```
akbashev, line 138: swift format change

PR #1236, opened 2026-05-27
Fix build given dependency changes

ktoso:
Fixes the build given changes in dependencies

resolves rdar://177615113


Sources/DistributedCluster/Cluster/Downing/DowningStrategy.swift
```diff
@@ -147,7 +147,7 @@ internal distributed actor DowningStrategyShell {
                 metadata: self.metadata([
                     "downing/node": "\(reflecting: member.node)",
```
kukushechkin: I would recommend to keep using the `extension String.StringInterpolation` (not adding `extension Logger.MetadataValue.StringInterpolation`), and change the emission to explicitly create a `MetadataValue` from a `String`:

```suggestion
                    "downing/node": .string("\(reflecting: member.node)"),
```

ktoso: Okey let me do that

ktoso:
nightly has TBD validation issues which ill check separately; for now merging the build fixes

PR #1237, opened 2026-05-28
Fix warnings

akbashev:
Going through a wall of compiler warnings every time we check the build makes it easy to miss real problems. 

This PR fixes the majority of those warnings in the core library and test runner so we can have a clean, easy-to-read build output.


ktoso (approved):
Thank you, looks good

ktoso:
ignore nightly for now

PR #1239, opened 2026-06-03
Membership without protocols - 0.4.0 dependency update

akbashev:
Update ClusterSystem if [PR in swift-cluster-membership](https://github.com/apple/swift-cluster-membership/pull/139) will approved:
- Replace generic SWIM.Instance<SWIMActor,…> → SWIM.Instance
- Remove SWIMActor peer protocol conformances (SWIMPeer, SWIMPingOriginPeer, SWIMPingRequestOriginPeer)
- Change distributed ping/pingRequest methods to accept ClusterMembership.Node instead of SWIMActor
- Update protobuf schema: ActorID → ClusterNode for members and ping responses
- Fix Node deserialisation: read name from proto.system instead of proto.protocol


Package.swift
```diff
@@ -189,7 +189,8 @@ var dependencies: [Package.Dependency] = [
 
     // .package(url: "https://github.com/apple/swift-cluster-membership", from: "0.3.0"),
     //    .package(name: "swift-cluster-membership", path: "Packages/swift-cluster-membership"), // FIXME: just work in progress
-    .package(url: "https://github.com/apple/swift-cluster-membership", branch: "main"),
+    // FIXME: revert changes back when swift-cluster-membership is merged
```
akbashev: This should be reverted when https://github.com/apple/swift-cluster-membership/pull/139 will land

Sources/DistributedCluster/Cluster/SWIM/SWIMActor+NotUsedCalls.swift
```diff
@@ -1,35 +0,0 @@
...
-import ClusterMembership
-import CoreMetrics
-import SWIM
-
-import struct Dispatch.DispatchTime
-import enum Dispatch.DispatchTimeInterval
-
-extension SWIMActor: SWIMPingOriginPeer {
-    nonisolated func ack(acknowledging sequenceNumber: SWIM.SequenceNumber, target: SWIMActor, incarnation: SWIM.Incarnation, payload: SWIM.GossipPayload<SWIMActor>) async throws {
-        fatalError("SWIM.ACKs are sent directly as replies to ping requests; the \(#function) is not used or implemented in ClusterSystem")
```
ktoso, line 24: Yeah that's a good reason to kill off the protocol I suppose, lgtm

Package.swift
ktoso: Done, And I'll follow up on this change

ktoso (approved):
This does simplify the setup a bit and bumps us to 0.4 so it's a good step, thanks for the cleanup.

I'm unsure about the actor id to node move, technically an id is unique to connection but node isnt... but deathwatch should have taken care of cleanups like these hmmm

